PR #1, opened 2015-04-28
Add bad_route page as redirect for invalid routes to Cerner's FHIR Serve...

dennispatterson:
...r.

When new users are trying Cerner's FHIR server url and may navigate to the tenant/application root in their browsers, this bad_route page serves as a redirect location.  The change under review migrates the highlights div to its own layout so that the homepage and the bad_route page are essentially the same, but with a different message.

To view:
1. Checkout the ServerRedirectPage branch
2. bundle install
3. run nanoc to generate static content
4. run 'nanoc view' and hit http://localhost:3000/bad_route/ to view page

Separate pull request will be sent for generated changes to gh-pages branch.

### DIFF
--- a/content/bad_route.md
+++ b/content/bad_route.md
@@ -1,0 +1,14 @@
+---
+title: Cerner | Bad Route
+layout: overview
+---
+
+<div class="wrapper feature">
+  <h1>
+    We notice you're trying to call our FHIR<br/>
+    server but your request is to a bad route.
+  </h1>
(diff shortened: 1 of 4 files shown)


kpshek:
I'm holding off on this change until we figure out the rewrite rule

daliboz:
+1, will use meta tags to redirect for now

kpshek:
:+1:

PR #7, opened 2015-05-08
Add DSTU2 for allergies and immunization

naveenaluru:
Adding DSTU2 documentation for allergies and immunization


jfrhong:
+1

dennispatterson:
:+1: after squash

PR #12, opened 2015-07-01
Initial commit of the authorization client-facing documentation

icon5585:
Creating this pull request as an internal github review for the authorization documentation.

### DIFF
--- a/content/dstu2/Authorization.md
+++ b/content/dstu2/Authorization.md
@@ -32,7 +32,7 @@ Once a launch context is received by your client application from the EHR, it mu
 
 * response_type
 * client_id
-* launch - Note: only required if using [SMART][5] launch
+* launch - Note: only required if using a [SMART][5] launch context
 * scope - Note: launch **must** be one of the scopes if using a [SMART][5] launch context
 
 It is also **recommended** that your client application provides the following query parameters.
--- a/content/dstu2/Authorization.md
+++ b/content/dstu2/Authorization.md
@@ -1,0 +1,154 @@
+# Authorization Server Client Documentation #
+---------------------------------------------
+
+### Introduction ###
+The Cerner Authorization Server currently supports OAuth 2.0 [1] SMART [2, 3] on FHIR launch workflows. As a client, it will require interaction between the client (you), the user, the authorization server, a SMART Launch server and a FHIR resource server.
+
+### Registration ###
+In order for your client application to utilize any FHIR protected resources, you must first register your client application. To do this, you can contact your Cerner representative with the following information, the following fields are **required**.
+
+* Name
+* Redirect/Callback URI
+* Email address
+
+You may also provide the following **optional** fields.
+
+* Logo URI 
+* SMART Launch Server URI
+
+Once approved, a **client identifier** will be provided to you for use with the Cerner Authorization Server. As a registered client, Cerner tenants may then ask for your client to be enabled, which is necessary in order to gain access to their protected FHIR resources.
+
+### Supported Scopes ###
+The Cerner Authorization Server supports many, but not all of the SMART or OAuth scopes. The following scopes are supported.
+
+* online_access
+* launch
+* openid
+* profile
+
+### Requesting an authorization code ###
+Cerner currently supports the SMART launch workflow from within an EHR, such as Cerner Millennium's PowerChart. This allows for current context information (patient info, encounter info, user info, etc.) to be provided to the client application upon launching. Below is a flowchart of the EHR SMART launch workflow.
+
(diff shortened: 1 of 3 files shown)


kpshek:
Rather than provide a list of references at the bottom of Authorization.md, let's just link the references inline so the reader can follow the reference link immediately.

kpshek:
References to "SMART Launch server" are likely to confuse developers since this is an internal implementation detail in our system. Others implementing SMART may bake this directly into the EHR app rather than the approach we've taken. I suggest we just omit all references to this and instead use 'EHR' which is what the SMART documentation uses.

content/dstu2/Authorization.md
```diff
@@ -0,0 +1,156 @@
...
+### Registration ###
+In order for your client application to utilize any FHIR protected resources, your client application must first register. To do this, contact your Cerner representative with the following information, the following fields are **required**.
+
+* Name - Application name, this should match what the name of the app will be in the app store
+* Redirect/Callback URI
+* Email address
+
+You may also provide the following **optional** fields.
+
+* Logo URI
```
kpshek: I think we should go ahead and make this a required field now since we're going to in the future. Additionally, we need further details as to the dimensions of this logo.

content/dstu2/Authorization.md
```diff
@@ -0,0 +1,156 @@
...
+In order for your client application to utilize any FHIR protected resources, your client application must first register. To do this, contact your Cerner representative with the following information, the following fields are **required**.
+
+* Name - Application name, this should match what the name of the app will be in the app store
+* Redirect/Callback URI
+* Email address
+
+You may also provide the following **optional** fields.
+
+* Logo URI 
+* SMART Launch Server URI
```
kpshek: What is this?

content/dstu2/Authorization.md
```diff
@@ -0,0 +1,156 @@
+# Authorization Server Client Documentation #
+---------------------------------------------
+
+### Introduction ###
+The Cerner Authorization Server currently supports OAuth 2.0 [1] SMART [2, 3] on FHIR launch workflows. As a client, it will require interaction between the client (your client application), the user, the authorization server, a SMART Launch server and a FHIR resource server.
```
kpshek: References to "SMART Launch server" are likely to confuse developers since this is an internal implementation detail in our system. Others implementing SMART may bake this directly into the EHR app rather than the approach we've taken. I suggest we just omit all references to this and instead use 'EHR' which is what the SMART documentation uses.

content/dstu2/Authorization.md
```diff
@@ -0,0 +1,156 @@
...
+
+Once a launch context is received by your client application from the SMART launch server, it must be sent to the authorization server so that an authorization code may be requested. To request an authorization code, to turn in for an access code, you will need to issue a **GET** request to the authorization server's authorize endpoint. At this point the authorization server will redirect the user to authenticate if they are not already authenticated. Your client application should use the system's browser for performing this exchange rather than an embedded browser. Using embedded browsers prevents single sign-on and authentication may fail for certain organizations who implement third party authentication systems.
+
+The authorize endpoints are tenant specific, which means your client application will need to know the specific tenant ID of the tenants whose data your client application is going to request access to as well as the following **required** query parameters.
```
kpshek: I don't believe clients will need to know the tenant id (if they do today, we need to fix that).

content/dstu2/Authorization.md
whitehatguy: I'm of the opinion we should simply require an SVG for the logo.  This will reduce any pains with scaling and presentation across a wide range of devices and concerns around exact dimensions.  By the time this goes mainstream, IE8 should drop from its current 2% of traffic to less than 1%, so SVG is going to be well supported.

content/dstu2/Authorization.md
icon5585: This has been removed

content/dstu2/Authorization.md
icon5585: Will make it required, +1 to SVG

icon5585:
Will do inline references

content/dstu2/Authorization.md
icon5585: If they're doing SMART launch workflows, the authorize endpoint will come from the metadata doc. I will document as such in here

content/dstu2/Authorization.md
icon5585: updated to EHR

daliboz:
We probably should update the [main page for DSTU 2] (https://github.com/cerner/fhir.cerner.com/blob/Authorization/content/dstu2.md#authentication) to reference this section (and re-title it)

icon5585:
Updated main DSTU2 page to reference this doc and renamed to Authorization

icon5585:
You are correct that if they're discovering it via FHIR the URL is an implementation detail, however, I'd like to provide some examples as close to real-world as possible in this doc. I can remove the {TENANT_ID} part and just put some placeholder text if that would help?

daliboz:
+1

icon5585:
Everything look good @kpshek ? :)

PR #18, opened 2015-10-09
update allergyIntolerance with new status param, new examples

daliboz:
Updated params. Added a required field (will be pushing this through to other resources). Thoughts on doing this this way?

Updated examples.

### DIFF
--- a/lib/resources.rb
+++ b/lib/resources.rb
@@ -1300,11 +1300,11 @@ module Cerner
         "resourceType"=> "Bundle",
         "id"=> "5ab17194-bb85-45fe-94d5-a02a2e8699ab",
         "type"=> "searchset",
-        "base"=> "https=>//fhir.sandboxcernerpowerchart.com/may2015/open/d075cf8b-3261-481d-97e5-ba6c48d3b41f",
+        "base"=> "https://fhir.sandboxcernerpowerchart.com/may2015/open/d075cf8b-3261-481d-97e5-ba6c48d3b41f",
         "total"=> 2,
         "link"=> [{
                       "relation"=> "self",
-                      "url"=> "https=>//fhir.sandboxcernerpowerchart.com/may2015/open/d075cf8b-3261-481d-97e5-ba6c48d3b41f/AllergyIntolerance?patient=2744010"
+                      "url"=> "https://fhir.sandboxcernerpowerchart.com/may2015/open/d075cf8b-3261-481d-97e5-ba6c48d3b41f/AllergyIntolerance?patient=2744010"
                   }],
         "entry"=> [{
                        "resource"=> {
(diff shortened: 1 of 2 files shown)


lib/resources.rb
```diff
@@ -1297,50 +1297,83 @@ def get_resource(key)
...
+                      "relation"=> "self",
+                      "url"=> "https=>//fhir.sandboxcernerpowerchart.com/may2015/open/d075cf8b-3261-481d-97e5-ba6c48d3b41f/AllergyIntolerance?patient=2744010"
+                  }],
+        "entry"=> [{
+                       "resource"=> {
+                           "resourceType"=> "AllergyIntolerance",
+                           "id"=> "3643731",
+                           "meta"=> {
+                               "versionId"=> "3643735",
+                               "lastUpdated"=> "2015-03-11T20=>34=>59.000Z"
```
MaxPhilips: Switch back to colons, lines 1303, 1307, 1321, 1332, 1349, 1355, 1366 too

lib/resources.rb
```diff
@@ -1297,50 +1297,83 @@ def get_resource(key)
...
+                           },
+                           "text"=> {
+                               "status"=> "generated",
+                               "div"=> "&lt;div>&lt;p>&lt;b>Allergy Intolerance&lt;/b>&lt;/p>&lt;p>&lt;b>Patient&lt;/b>: RHEUM, TEST ONE&lt;/p>&lt;p>&lt;b>Allergy&lt;/b>: Penicillin&lt;/p>&lt;p>&lt;b>Status&lt;/b>: Refuted&lt;/p>&lt;p>&lt;b>Criticality&lt;/b>: &lt;/p>&lt;p>&lt;b>Category&lt;/b>: Medication&lt;/p>&lt;p>&lt;b>Reactions&lt;/b>: &lt;/p>&lt;p>&lt;b>Onset&lt;/b>: &lt;/p>&lt;p>&lt;b>Comment&lt;/b>: &lt;/p>&lt;/div>"
```
MaxPhilips: Looks fine on the site itself but do we want to either write the &lt;'s as <'s or the >'s as &gt;'s?

MaxPhilips:
I think the required column is definitely helpful.  Although it looks a little strange to have the Y/N's as normal text while the entries in Name and Type are monospaced.  Does it make sense to monospace the Required? entries just to keep things looking similar?

MaxPhilips:
+1 after resources.rb updates

lib/resources.rb
daliboz: I kept only the < as &lt; so that it wouldn't display the text as raw html

daliboz:
Corrected hash rocket issues.

lib/resources.rb
```diff
@@ -1297,50 +1297,83 @@ def get_resource(key)
...
+                               "lastUpdated"=> "2015-03-11T20:34:59.000Z"
+                           },
+                           "text"=> {
+                               "status"=> "generated",
+                               "div"=> "&lt;div>&lt;p>&lt;b>Allergy Intolerance&lt;/b>&lt;/p>&lt;p>&lt;b>Patient&lt;/b>: RHEUM, TEST ONE&lt;/p>&lt;p>&lt;b>Allergy&lt;/b>: Penicillin&lt;/p>&lt;p>&lt;b>Status&lt;/b>: Refuted&lt;/p>&lt;p>&lt;b>Criticality&lt;/b>: &lt;/p>&lt;p>&lt;b>Category&lt;/b>: Medication&lt;/p>&lt;p>&lt;b>Reactions&lt;/b>: &lt;/p>&lt;p>&lt;b>Onset&lt;/b>: &lt;/p>&lt;p>&lt;b>Comment&lt;/b>: &lt;/p>&lt;/div>"
+                           },
+                           "recordedDate"=> "2015-03-11T15:34:59.000",
+                           "recorder"=> {
+                               "reference"=> "Practitioner/3270007",
+                               "display"=> "Venkat, Savitha"
```
dennispatterson: I realize that her name will come back when the url is queried, but any concern over publishing Savitha's name on here?

jfrhong:
+1

daliboz:
Updated the display name to be more example-like.

dennispatterson:
+tardis

PR #22, opened 2015-11-02
Update DiagnosticReport documentation and response

daliboz:
Updates for the new May 2015 response fields and required parameter documentation

### DIFF
--- a/lib/resources.rb
+++ b/lib/resources.rb
@@ -1453,27 +1453,163 @@ module Cerner
          }
 
     MAY2015_DIAGNOSTIC_REPORT_BUNDLE ||= {
-        "resourceType"=>"Bundle",
-        "id"=> "b2502416-98eb-4a53-b9cc-3262c1617a71",
-        "type"=> "searchset",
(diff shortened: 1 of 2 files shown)


lib/resources.rb
```diff
@@ -1453,27 +1453,163 @@ def get_resource(key)
...
+                            {
+                                "system" => "http://hl7.org/fhir/v2/0074",
+                                "code" => "RAD",
+                                "display" => "Radiology"
+                            }
+                        ],
+                        "text" => "Radiology"
+                    },
+                    "diagnosticDateTime" => "2015-08-06T20:03:09.000Z",
+                    "conclusion" => "Report_asdfasdfasdf jjimmmy   Signature Line:***** Preliminary Report *****   Transcribed by: DW      Reason for exam:4061979"
```
namotron: Since this is going to be publicly visible, should we clean up some of these values?

cerbs: Since this is going to be publicly visible, should we clean up some of these values?

MaxPhilips:
+1

jfrhong:
+1

daliboz:
Updated the one diagnostic report with weird text.

dennispatterson:
:+1:

PR #23, opened 2015-11-02
Update Encounter with new response and parameters

daliboz:
Updates for May response fields and new parameters that were supported with the release.

### DIFF
--- a/content/may2015/encounter.md
+++ b/content/may2015/encounter.md
@@ -15,9 +15,10 @@ Search for Encounters that meet supplied query parameters:
 
 ### Parameters
 
- Name    | Type                                                           | Description
----------|----------------------------------------------------------------|-------------------------------------------------------
-`patient`|[`reference`](http://hl7.org/fhir/2015May/search.html#reference)| The patient present at the encounter. Example: `12345`
+ Name    | Required? | Type                                                           | Description
+---------|-----------|----------------------------------------------------------------|-------------------------------------------------------
+`patient`| This or _id |[`reference`](http://hl7.org/fhir/2015May/search.html#reference)| The patient present at the encounter. Example: `12345`
+`_id`    | This or patient |[`id`](http://hl7.org/fhir/2015May/datatypes.html#id) | A single or comma separated list of Encounter ids. Example: `45678,89765` 
 
 ### Response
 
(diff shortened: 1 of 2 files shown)


dennispatterson:
:+1:

content/may2015/encounter.md
```diff
@@ -15,9 +15,10 @@ Search for Encounters that meet supplied query parameters:
 
 ### Parameters
 
- Name    | Type                                                           | Description
----------|----------------------------------------------------------------|-------------------------------------------------------
-`patient`|[`reference`](http://hl7.org/fhir/2015May/search.html#reference)| The patient present at the encounter. Example: `12345`
+ Name    | Required? | Type                                                           | Description
+---------|-----------|----------------------------------------------------------------|-------------------------------------------------------
+`patient`| This or _id |[`reference`](http://hl7.org/fhir/2015May/search.html#reference)| The patient present at the encounter. Example: `12345`
+`_id`    | This or patient |[`id`](http://hl7.org/fhir/2015May/datatypes.html#id) | A single or comma separated list of Encounter ids. Example: `45678,89765`
```
MaxPhilips: Change description to "The Encounter id, may be a list separated by commas. Example: `45678,89765`" to avoid the id/ids problem?

lib/resources.rb
```diff
@@ -1477,37 +1477,181 @@ def get_resource(key)
...
+        "base" => "https://fhir-open.sandboxcernerpowerchart.com/may2015/d075cf8b-3261-481d-97e5-ba6c48d3b41f",
+        "total" => 2,
+        "link" => [
+            {
+                "relation" => "self",
+                "url" => "https://fhir-open.sandboxcernerpowerchart.com/may2015/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Encounter?_id=3261906%2C2787906"
+            }
+        ],
+        "entry" => [
+            {
```
MaxPhilips: We should still be able to use a reference to MAY2015_ENCOUNTER for the first resource to eliminate duplication.

MaxPhilips:
+1

jfrhong:
+1

lib/resources.rb
daliboz: Not sure it's really worth it - I don't actually require the id lookup to be the same one as one of the search ones...

content/may2015/encounter.md
daliboz: I don't think this reads poorly. I could put a () around the s if this is an issue?

PR #26, opened 2015-11-12
Update Observation for May DSTU ballot

daliboz:
Add additional parameters for Observation, update response example with new fields and paging


jfrhong:
+1

dennispatterson:
:+1:

PR #32, opened 2016-01-06
Add initial documentation for DocumentReference DSTU2

cerbs:


lib/definition_table_generator.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def field_definition_table(content_key, version)
+
+    schema = YAML.load(File.read("lib/resources/#{version.to_s}/#{content_key}.yaml"))
+
+    # TODO: This concept needs to be reworked to allow multiple types for a given resource, or multiple versions of
+    # the same type (e.g. Reference(Patient|Practitioner|RelatedPerson)). It would be nice if it also did not apply the
+    # html formatting; that should be done elsewhere.
+    types = YAML.load(File.read("lib/resources/#{version.to_s}/types.yaml"))
+
```
cerbs: Extra lines can be removed

lib/resources/dstu2/document_reference_create.yaml
```diff
@@ -0,0 +1,129 @@
...
+- name: author
+  required: 'Yes'
+  cardinality: 0..1
+  description: Who and/or what authored the document
+  example: '{"author": [ {"reference": "Practitioner/2150097" } ]}'
+
+- name: indexed
+  required: 'Yes'
+  cardinality: 1..1
+  type: dateTime
```
cerbs: indexed is an 'instant'

lib/resources/dstu2/document_reference_create.yaml
```diff
@@ -0,0 +1,129 @@
...
+  description: Kind of document.
+  example: |
+             {"coding": [
+              {"system": "http://loinc.org",
+               "code": "34840-9"}]}
+  note: The type must include a loinc coding.
+
+- name: author
+  required: 'Yes'
+  cardinality: 0..1
```
cerbs: Add type

lib/resources/dstu2/document_reference_create.yaml
```diff
@@ -0,0 +1,130 @@
+---
+field_name_base_url: http://hl7.org/fhir/DSTU2/documentreference-definitions.html#DocumentReference
+fields:
+- name: resourceType
```
daliboz: The link to the resourceType (and other fields defined as part of the core resource or buried in the referenced type) don't work since the links all assume they're on the DocumentReference page.

lib/resources/dstu2/document_reference_create.yaml
```diff
@@ -0,0 +1,130 @@
...
+      cardinality: 0..1
+      type: string
+      description: Data inline, base64-ed
+      example: |
+                {"content": [
+                  {"attachment": {
+                    "data": "PCFET0NUWVBFIGh0bWwNCiAgU1lTVEVNI...snipped"}}]}'
+      url: http://hl7.org/fhir/DSTU2/datatypes-definitions.html#Attachment.data
+
+- name: context
```
daliboz: Missing cardinality on context (0..1)

content/dstu2/document-reference.md
```diff
@@ -0,0 +1,69 @@
...
+---
+
+# DocumentReference
+
+* TOC
+{:toc}
+
+## Create
+
+Create and store a new clinical note document.
```
daliboz: "Create new documents. Only documents that would be classified as notes currently supported." ?

lib/resources/dstu2/document_reference_create.yaml
```diff
@@ -0,0 +1,130 @@
+---
```
daliboz: General question to all: the cardinality shown here is directly from the HL7 website. Should these instead reflect our profiled cardinality? IE: 1..1 or 1..* for required?

daliboz:
+1 other than comments above.

lib/resources/dstu2/document_reference_create.yaml
cerbs: I would be concerned that tracking cardinality values different than the official spec would end up causing more confusion and we'd have to fend off attempts to 'correct' the massaged cardinalities both from our future-selves and well-intentioned consumers. It won't be clear without actually looking at the code if we have intentionally changed the cardinality due to implementation, or if it is in fact just a typo. I hesitated to list the cardinalities due to this reason. 

Another tact is to consider what they actually mean to us in practice; whether the field is optional or required, and whether it is a list.  The Required column already implies the cardinality of a single valued field. We could just replace the cardinality cell with whether it is a list or not.

lib/resources/dstu2/document_reference_create.yaml
cerbs: A couple things going on here:

1. The resourceType link was basically working as intended to take you to the top level of the DocumentReference field list, but it should be linking to the resourceType definition instead, so I have added the url for the field to point to the correct location.

2. The url generation logic was incorrectly flipped during some refactoring so that the override urls would only be used if there was no base url defined. I rearranged that, so now the overrides should link properly for the relevant fields (resourceType, content.attachment.contentType, content.attachment.data)

lib/resources/dstu2/document_reference_create.yaml
```diff
@@ -0,0 +1,132 @@
...
+              {"system": "http://loinc.org",
+               "code": "34840-9"}]}
+  note: The type must include a loinc coding.
+
+- name: author
+  required: 'Yes'
+  cardinality: 0..1
+  type: PractitionerReference
+  description: Who and/or what authored the document
+  example: '{"author": [ {"reference": "Practitioner/2150097" } ]}'
```
MaxPhilips: Nit: consistently format the spaces between brackets from example to example

lib/resources/dstu2/document_reference_create.yaml
MaxPhilips: +1 to matching the official spec.  If we use different values we should add a note explaining the difference

MaxPhilips:
+1

cerbs:
Rebasing and squashing

PR #36, opened 2016-01-13
Adding AllergyIntolerance DSTU2 documentation

cerbs:
Changes:
- Adding AllergyIntolerance DSTU2 documentation for Search, Create, and Update
- Enhancing the definition_table logic
    - Adding support for flagging fields and values for different action types
    - Adding support for simple embedding linking in notes and description fields
    - Refactoring type field definition and linking logic to be more straightforward and flexible
    - Adding 'top' link to each entry to return to the top of the action
- Updating README.md with documentation on how to use and config the definition_table method

Things to look at when running locally:
- Verify that AllergyIntolerance documentation looks accurate
- Because the type linking was reworked the field names and type links should be looked at again in may2015 and DSTU2. It shouldn't be necessary to verify DocReference AND AllergyIntolerance since they share the same system. Spot checks should be good.
- In AllergyIntolerance Update table:
  - Verify that Practitioner in the recorder field links to the FHIR Practitioner documentation
  - Verify that 'contained' in the reporter notes links to the contained field in the Update table. 
  - Verify that any of the 'top' links take you to the top of the Update section. Any of the 'top' links in the Create section should likewise take you to the top of Create.

### DIFF
--- a/content/dstu2/allergy-intolerance.md
+++ b/content/dstu2/allergy-intolerance.md
@@ -1,0 +1,130 @@
+---
+title: AllergyIntolerance | FHIR DSTU 2 API
+---
+
+# AllergyIntolerance
+
+* TOC
+{:toc}
+
+## Search
+
+Search for AllergyIntolerances that meet supplied query parameters:
+
+    GET /AllergyIntolerance?:parameters
+
+### Parameters
+
+ Name    | Required? | Type                                                           | Description
+---------|-----------|----------------------------------------------------------------|---------------------------------------------
+`patient`|     Y     |[`reference`](http://hl7.org/fhir/DSTU2/search.html#reference)| Who the sensitivity is for. Example: `12345`
+`status` |     N     |[`token`](http://hl7.org/fhir/DSTU2/search.html#token)        | [Certainty of the allergy or intolerance](http://hl7.org/fhir/DSTU2/valueset-allergy-intolerance-status.html). Example: `confirmed`
+
+### Response
+
+<%= headers 200 %>
+<%= json(:dstu2_allergy_intolerance_bundle) %>
+
+## Create
+
+Create new allergies.
+
+    POST /AllergyIntolerance
+
+### Headers
+
+To successfully POST an allergy, the following headers must be provided. Allergy creation is supported from the closed endpoint only.
+
+    Content-Type: application/json+fhir
+    Authorization: <OAuth2 Bearer Token>
+
+### Body fields
+
+<%= definition_table(:allergy_intolerance, :create, :dstu2) %>
+
+### Example Body
+
+<%= json(:dstu2_allergy_intolerance_create) %>
+
+### Response
+
+<%= headers 201 %>
+<pre class="terminal">
+    Connection → Keep-Alive
+    Content-Encoding → gzip
+    Content-Length → 20
+    Content-Type → text/html; charset=UTF-8
+    Date → Wed, 13 Jan 2016 21:45:47 GMT
+    Keep-Alive → timeout=15, max=100
+    Last-Modified → Tue, 15 Dec 2015 19:13:20 GMT
+    Status → 201 Created
+    access-control-allow-methods → DELETE, GET, POST, PUT, OPTIONS, HEAD
+    access-control-allow-origin → *
+    access-control-expose-headers → ETag, Content-Location, Location, X-Request-Id, WWW-Authenticate, Date
+    access-control-max-age → 0
+    cache-control → no-cache
+    etag → W/"35502191"
+    location → https://fhir.sandboxcernerpowerchart.com/dstu2/9e2aaa6d-3811-4d84-b5f9-93ccf529eefa/AllergyIntolerance/35502191
+    server-response-time → 1260.984596
+    strict-transport-security → max-age=631152000
+    vary → Origin,User-Agent,Accept-Encoding
+    x-content-type-options → nosniff
+    x-frame-options → SAMEORIGIN
+    x-request-id → edad5615-1f19-4ff0-9999-46f46a52368e
+    x-runtime → 1.260940
+    x-xss-protection → 1; mode=block
+</pre>
+
+The `etag` response header indicates the current `If-Match` version to use on subsequent updates.
+
+## Update
+
+Update an existing allergy.
+
+    PUT /AllergyIntolerance
+
+Note that any field which is missing will be interpreted as nulling out or removing data from the resource. See [FHIR Update](http://hl7.org/fhir/DSTU2/http.html#update) for additional details about update operations.
+
+### Headers
+
+To successfully PUT an allergy, the following headers must be provided. Allergy updates are supported from the closed endpoint only.
+
+    Content-Type: application/json+fhir
+    Authorization: <OAuth2 Bearer Token>
+    If-Match: W/"<Current allergy id>"
+
+### Body fields
+
+<%= definition_table(:allergy_intolerance, :update, :dstu2) %>
+
+### Example Body
+
+<%= json(:dstu2_allergy_intolerance_update) %>
+
+### Response
+
+<%= headers 200 %>
+<pre class="terminal">
+    Connection → Keep-Alive
+    Content-Encoding → gzip
+    Content-Length → 20
+    Content-Type → text/html; charset=UTF-8
+    Date → Wed, 13 Jan 2016 21:50:53 GMT
+    Keep-Alive → timeout=15, max=100
+    Last-Modified → Tue, 15 Dec 2015 19:13:20 GMT
+    Status → 200 OK
+    access-control-allow-methods → DELETE, GET, POST, PUT, OPTIONS, HEAD
+    access-control-allow-origin → *
+    access-control-expose-headers → ETag, Content-Location, Location, X-Request-Id, WWW-Authenticate, Date
+    access-control-max-age → 0
+    cache-control → no-cache
+    etag → W/"35504311"
+    server-response-time → 653.7616069999999
+    strict-transport-security → max-age=631152000
+    vary → Origin,User-Agent,Accept-Encoding
+    x-content-type-options → nosniff
+    x-frame-options → SAMEORIGIN
+    x-request-id → 9dba8326-899a-406f-a125-3fc3d6605dad
+    x-runtime → 0.653722
+    x-xss-protection → 1; mode=block
+</pre>
--- a/lib/field_definition_table.erb
+++ b/lib/field_definition_table.erb
@@ -9,44 +9,39 @@
   </thead>
   <tbody>
     <% for item in data[:fields] %>
-
-      <!-- TODO: Externalize the styling in this erb so that it is consistent with the rest of the site. Keeping in
-                 mind that the table as a whole needs to not scroll. Individual examples should, but the entire table
-                 should not. It looks tacky -->
-      <tr>
-        <td>
-          <!-- Render the field name as a link, if a url is provided.-->
-          <% if !item[:url].nil? %>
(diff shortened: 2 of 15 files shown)


lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -0,0 +1,211 @@
...
+  required:
+  - update: 'Yes'
+  - create: <b>Invalid</b>
+  cardinality: 0..1
+  type: id
+  description: The logical id of the resource to update.
+  example: '{"id": "123412"}'
+  note:
+  - update: The id value must match the AllergyIntolerance/&lt;id> value.
+  - create: The id field <b>must not</b> be set when performing an update operation.
```
MaxPhilips: I know we talked about this a bit earlier.  After seeing this in the context of all the other fields, it seemed a lot more confusing to me than before when I thought we'd want to include it.  Maybe we want to convey this information in a note above/below the table instead of a row?

lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -0,0 +1,211 @@
...
+  cardinality: 0..1
+  type: Reference (Patient) | contained Reference (Practitioner | RelatedPerson)
+  description: Source of the information about the allergy.
+  example: |
+             Reporter as a Patient Reference
+               {"reporter": {
+                  "reference": "Patient/5366327"}}
+
+             Reporter as a contained reference
+               {"reporter": "#5366327}
```
MaxPhilips: I'd say change this id and the one in the update table to match the example in contained - 123

lib/field_definition_table.erb
```diff
@@ -9,44 +9,39 @@
...
+      <% unless item[:example].nil? %>
+        <li><i>Example</i></li>
+        <ul style="list-style-type: none"><li><pre><%= item[:example] %></pre></li></ul>
+      <%end%>
+      <% unless item[:note].nil? %>
+        <li><i>Notes</i></li>
+        <ul style="list-style-type: none"><li><%= item[:note] %></li></ul>
+      <%end%>
+      </ul>
+      <div style="float: right"><a href="#<%=data[:table_name]%>">top</a></div>
```
MaxPhilips: * Is this necessary? Another hyperlink in each table cell looks like something I want to click to learn more.  The current placement (only right-aligned word) also seems to break the flow a bit
* Is there a better name that conveys 'top of table'? At first I thought this would go to the top of the page

content/dstu2/allergy-intolerance.md
```diff
@@ -0,0 +1,128 @@
...
+
+ Name    | Required? | Type                                                           | Description
+---------|-----------|----------------------------------------------------------------|---------------------------------------------
+`patient`|     Y     |[`reference`](http://hl7.org/fhir/DSTU2/search.html#reference)| Who the sensitivity is for. Example: `12345`
+`status` |     N     |[`token`](http://hl7.org/fhir/DSTU2/search.html#token)        | [Certainty of the allergy or intolerance](http://hl7.org/fhir/DSTU2/valueset-allergy-intolerance-status.html). Example: `confirmed`
+
+### Response
+
+<%= headers 200 %>
+<%= json(:dstu2_allergy_intolerance_bundle) %>
```
MaxPhilips: JSON rendered this way is used in create and update as the request body style instead of the response body.  We might want to provide a visual queue in addition to the section headers to differentiate request vs response.  If we do change this it will require changes in the other resources too

lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -0,0 +1,211 @@
...
+  type: Resource
+  description: Contained, inline Resources.
+  example: |
+             {"contained": [
+                {"resourceType": "RelatedPerson",
+                 "id": "123",
+                 "relationship": {
+                    "coding": [
+                      {"system": "http://hl7.org/fhir/v3/RoleCode",
+                       "code": "SIGOTHR"}]}}]}
```
MaxPhilips: May want to add an example of a contained Practitioner.
```
{"contained": [
   {"resourceType": "Practitioner",
    "id": "123",
    "practitionerRole": [
      {"role": {
         "coding": [
           {"system": "http://hl7.org/fhir/v2/0286",
             "code": "RP"}]}}]}]
```

lib/field_definition_table.erb
cerbs: I can rename to 'top of section' but I do think it is necessary or at the least, desirable. I didn't add it until I was consistently getting annoyed when trying to actually use my own documentation and kept getting lost, especially after adding the Update section when suddenly I had 2 nearly identical tables and I wasn't sure which version of the table I was looking at.  

Ultimately I think it'd be better to use some scripting and render just one table with the ability to toggle between actions, but who has time for that? :)  And even then, with a sufficiently long table it would still be nice to get back up to the top quickly.

As for the placement of the text, since it isn't content it is intentionally out of the way as to be unobtrusive.

lib/resources/dstu2/allergy_intolerance.yaml
cerbs: I've switched this to only display in the update context. We'll rely on consumers to be familiar enough with the spec to know they shouldn't pass it in.

content/dstu2/allergy-intolerance.md
cerbs: I don't find it that confusing since the context is labeled. Unless others have strong opinions, I'll leave this as-is.

lib/resources/dstu2/allergy_intolerance.yaml
cerbs: I bumped this up to 5366328 and changed the other contained id. It looks weird to have both a believably valid id and an obviously fake id in the same example.

lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -0,0 +1,219 @@
...
+               {"contained": [
+                  {"resourceType": "Practitioner",
+                   "id": "123",
+                   "practitionerRole": [
+                     {"role": {
+                        "coding": [
+                          {"system": "http://hl7.org/fhir/v2/0286",
+                            "code": "RP"}]}}]}]
+  note: >
+          These should be either `Practitioner` or `RelatedPerson` resources that are referenced with the
```
daliboz: The Practitioner links to the correct place, but RelatedPerson attempts to link to a datatype instead of a Resource.

lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -0,0 +1,219 @@
...
+- name: id
+  required: 'Yes'
+  cardinality: 0..1
+  type: id
+  description: The logical id of the resource to update.
+  example: '{"id": "123412"}'
+  note: The id value must match the AllergyIntolerance/&lt;id> value.
+  action: update
+
+- name: contained
```
daliboz: Is this intended to be a link to the detailed allergy fields? Contained isn't one of them, so it just goes to the top of that resource. There is a section in the spec about contained resources: http://hl7.org/fhir/DSTU2/references.html#contained

lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -0,0 +1,219 @@
...
+  example: '{"onset: "2012-07-13T00:00:00Z"}'
+
+- name: recordedDate
+  required: 'Yes'
+  cardinality: 0..1
+  type: dateTime
+  description: When the sensitivity was recorded.
+  example: '{"recordedDate": "2015-10-14T13:13:20-06:00"}'
+
+- name: recorder
```
daliboz: Is recorder actually required? I thought that requirement was removed/it was a bug.

content/dstu2/allergy-intolerance.md
```diff
@@ -0,0 +1,128 @@
...
+    x-request-id → edad5615-1f19-4ff0-9999-46f46a52368e
+    x-runtime → 1.260940
+    x-xss-protection → 1; mode=block
+</pre>
+
+The `etag` response header indicates the current `If-Match` version to use on subsequent updates.
+
+## Update
+
+Update an existing allergy.
```
daliboz: We should add a disclaimer about any field that is missing is interpreted as nulling out/removing that field? Link to information about this in the spec: http://hl7.org/fhir/DSTU2/http.html#update

Or, update the general doc page around how updates behave and link to it from resources allowing updates?

content/dstu2/allergy-intolerance.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Search
+
+Search for AllergyIntolerances that meet supplied query parameters:
+
+    GET /AllergyIntolerance?:parameters
+
+### Parameters
+
+ Name    | Required? | Type                                                           | Description
```
daliboz: I think we need a new table, Terminology Bindings. Similar to: http://hl7.org/fhir/DSTU2/allergyintolerance.html#tx

But we would document the bindings for the non-required valuesets. This will eventually need to be throughout the resources, but it's definitely needed on the resources allowing write/update. Note that the binding doesn't change by action, so it can be towards the top/on it's own.

content/dstu2/allergy-intolerance.md
```diff
@@ -0,0 +1,128 @@
...
+---------|-----------|----------------------------------------------------------------|---------------------------------------------
+`patient`|     Y     |[`reference`](http://hl7.org/fhir/DSTU2/search.html#reference)| Who the sensitivity is for. Example: `12345`
+`status` |     N     |[`token`](http://hl7.org/fhir/DSTU2/search.html#token)        | [Certainty of the allergy or intolerance](http://hl7.org/fhir/DSTU2/valueset-allergy-intolerance-status.html). Example: `confirmed`
+
+### Response
+
+<%= headers 200 %>
+<%= json(:dstu2_allergy_intolerance_bundle) %>
+
+## Create
```
daliboz: Update the Client Errors section of the Overview page with new error codes to account for Add/Update. EG: 409, 422.

daliboz:
Other than previous comments, +1

lib/resources/dstu2/allergy_intolerance.yaml
cerbs: Cleaned up the type links and fixed RelatedPerson specifically:
https://github.com/cerner/fhir.cerner.com/commit/733a52e6680d4d66f1a62168e15df4eb4f48c112

lib/resources/dstu2/allergy_intolerance.yaml
cerbs: Updated: https://github.com/cerner/fhir.cerner.com/commit/c65d5b52943d9e5468bf04982486c32487e9c573#diff-f9143f9b065142644c34160aa818d5ddR67

content/dstu2/allergy-intolerance.md
cerbs: Added a disclaimer: https://github.com/cerner/fhir.cerner.com/commit/c65d5b52943d9e5468bf04982486c32487e9c573#diff-f3bcd4caf98ebce5446af6c4b939535cR86

content/dstu2/allergy-intolerance.md
cerbs: Added for may2015 and dstu2 https://github.com/cerner/fhir.cerner.com/commit/c975f197bab03b1ec5eaa0ec0f37f4efbfe63bba

content/dstu2/allergy-intolerance.md
cerbs: I have logged a separate issue to track these changes.

MaxPhilips:
+1

cerbs:
squashing

PR #40, opened 2016-02-03
Updates to help navigation with new styling

daliboz:
* Start Coding uses a local reference and navigates in same window/tab
* The "API" at the top of the page now states version so you know what you're looking at


jfrhong:
+1

cerbs:
:+1:

PR #46, opened 2016-02-09
Adding support for terminology binding table

cerbs:
Adding terminology table for supported terminologies of "Example" binding strength

### DIFF
--- a/lib/resources/dstu2/allergy_intolerance.yaml
+++ b/lib/resources/dstu2/allergy_intolerance.yaml
@@ -1,4 +1,5 @@
 ---
+name: AllergyIntolerance
 field_name_base_url: http://hl7.org/fhir/DSTU2/allergyintolerance-definitions.html#AllergyIntolerance
 fields:
 - name: resourceType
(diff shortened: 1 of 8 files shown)


lib/resources/dstu2/allergy_intolerance.yaml
```diff
@@ -109,6 +110,10 @@ fields:
                   {"system": "http://www.nlm.nih.gov/research/umls/rxnorm",
                    "code": "4125",
                    "display": "Ethiodized oil"}]}}
+  binding:
+    display: RxNorm
+    terminology: http://www.nlm.nih.gov/research/umls/rxnorm
```
daliboz: I believe we allow for snomed as well when the substance isn't a medication, and for negation.

daliboz:
Is the goal to document all terminologies for existing resources, or just a subset and then follow up with each one? If it's all, we're missing some.

cerbs:
A subset, specifically AllergyIntolerance and additionally DocumentReference because I'm familiar with it and it has actual known example values I could list to prove that functionality works. These changes were split off from the AllergyIntolerance Create/Update PR (https://github.com/cerner/fhir.cerner.com/pull/36) and is therefore focused there.

daliboz:
+1

MaxPhilips:
+1

PR #52, opened 2016-03-07
Add terminology bindings to DSTU 2 documentation for Condition resource.

naveenaluru:
Adding terminology table for condition resource

### DIFF
--- a/lib/resources/dstu2/condition.yaml
+++ b/lib/resources/dstu2/condition.yaml
@@ -73,7 +73,7 @@ fields:
                    "userSelected": true}],
                 "text": "Absence seizures"}}
   binding:
-    description: A code that defines the condition.
+    description: Identification of the condition or diagnosis.
     terminology:
     - display: SNOMED CT
       system: http://snomed.info/sct
@@ -95,7 +95,7 @@ fields:
                    "display": "Diagnosis"}],
                 "text": "Diagnosis"}}
   binding:
-    description: A code that defines the category of the condition.
+    description: A category assigned to the condition.
     terminology:
     - display: Condition Category Codes
       system: http://hl7.org/fhir/condition-category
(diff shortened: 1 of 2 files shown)


lib/resources/dstu2/condition.yaml
```diff
@@ -71,6 +72,15 @@ fields:
                    "display": "Generalized nonconvulsive epilepsy",
                    "userSelected": true}],
                 "text": "Absence seizures"}}
+  binding:
+    description: A code that defines the condition.
```
cerbs: These descriptions should probably describe the binding more than the value type. In other words, they should be closer to the values in the FHIR terminology table: http://hl7.org/fhir/DSTU2/condition.html#tx. In this case a lot of the terminology descriptions are the same, or nearly the same, as the field descriptions

lib/resources/dstu2/condition.yaml
```diff
@@ -71,6 +72,15 @@ fields:
...
+    terminology:
+    - display: SNOMED CT
+      system: http://snomed.info/sct
+    - display: ICD-9
+      system: http://hl7.org/fhir/sid/icd-9
```
cerbs: Is this the correct system for ICD-9?  The ICD-10 system url redirects to a Terminologies-systems page which lists all of the valid systems including that specific ICD-10 system, however it doesn't list this specific ICD-9 system. Instead it lists http://hl7.org/fhir/sid/icd-9-cm.  'http://hl7.org/fhir/sid/icd-9' does link to a FHIR page, but it looks like it is a working/temporary page. The page title literally has 'C:\work\org.hl7.fhir\...' in it.

lib/resources/dstu2/condition.yaml
naveenaluru: Done.

cerbs:
:+1:

lib/resources/dstu2/condition.yaml
naveenaluru: Agreed that 'http://hl7.org/fhir/sid/icd-9' is not listed in terminologies system. But this is currently what we are returning in response and the fhir page it links to provides source of this code system.

davischristi:
:+1:

PR #54, opened 2016-03-14
Return dispense quantity value as a number for MedicationPrescription/Order

dayouJ:


MaxPhilips:
Might want to update a few of the examples to be 20 or 20.52 as opposed to having them all as 20.0, assuming the other precisions can possibly occur.  It looks like all the examples are for tabs or boxes though, so it doesn't make sense to have 20.52 tabs.

+1 after

daliboz:
I think it's possible to have .5 tabs, even if .52 is unrealistic. +1 to suggestion of adding actual decimal.

dayouJ:
Added a couple example as quantity value with decimal.

namotron:
👍

cerbs:
+1 (The problem with having multiple github accounts is sometimes you realize later you were logged in with the wrong one)

PR #61, opened 2016-06-16
Adding DocumentReference implementation notes

cerbs:
- Adding "Implementation Notes" section to DocumentReference (may2015 and dstu2) to indicate that we do not support and ignore the relatesTo modifier
- Colonizing the DocumentReference.context.period.end example

### DIFF
--- a/content/dstu2/document-reference.md
+++ b/content/dstu2/document-reference.md
@@ -28,6 +28,10 @@ To successfully POST a document, the following headers must be provided. Documen
 
 <%= definition_table(:document_reference, :create, :dstu2) %>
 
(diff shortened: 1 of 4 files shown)


cerbs:
Wasn't sure the best way to represent implementation details like this so I decided to follow what the FHIR documentation does have just have an Implementation Notes section. It's quick and easy.

daliboz:
It feels like this should be more prominent - possibly immediately under the "body fields" and before the big table?

MaxPhilips:
+1

daliboz:
+1

PR #65, opened 2016-07-25
Update DSTU 2 documentation for Immunization.

jfrhong:
- Updated terminology bindings
- Updated example bundle

### DIFF
--- a/lib/dstu2_resources.rb
+++ b/lib/dstu2_resources.rb
@@ -361,104 +361,270 @@ module Cerner
         ]
     }
 
-    DSTU2_IMMUNIZATION_BUNDLE ||= {
-      "resourceType": "Bundle",
-      "id": "3639d63c-3475-4763-a98a-5f51d59d61bb",
-      "type": "searchset",
-      "total": 2,
-      "link": [
-        {
-          "relation": "self",
-          "url": "https://fhir-open.sandboxcernerpowerchart.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Immunization?patient=2652008"
-        }
-      ],
-      "entry": [
-        {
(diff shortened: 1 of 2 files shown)


lib/dstu2_resources.rb
```diff
@@ -361,104 +361,270 @@ module Resources
...
+              "id": "HM4234847",
+              "meta": {
+                "versionId": "0",
+                "lastUpdated": "2016-05-13T12:03:36-0500"
+              },
+              "text": {
+                "status": "generated",
+                "div": "&lt;div>&lt;p>&lt;b>Immunization&lt;/b>&lt;/p>&lt;p>&lt;b>Vaccine&lt;/b>: Hepatitis B&lt;/p>&lt;p>&lt;b>Date Attempted&lt;/b>: 2016-05-13T12:03:36-0500&lt;/p>&lt;p>&lt;b>Given&lt;/b>: No&lt;/p>&lt;p>&lt;b>Administered by&lt;/b>: Dr. Who&lt;/p>&lt;p>&lt;b>Reasons Not Given&lt;/b>: Patient Refuses&lt;/p>&lt;/div>"
+              },
+              "status": "on-hold",
```
daliboz: Is this still part of our returned value set? Per conversations around Argonaut, these statuses were removed.

daliboz:
Other than the question around the value set for status, +1.

lib/dstu2_resources.rb
jfrhong: We are not currently honor the Argonaut value set. There is a JIRA logged to switch to it.

cerbs:
+1, Other than the obligatory observation that 'Dr. Who' should be 'The Doctor'

PR #75, opened 2016-08-30
Clarify tenant terminology and wildcard scopes

daliboz:
* Updated reference to ehr source to just differentiate URLs change by tenant since this will be changing/can vary more than what's documented.
* Brought May and DSTU 2 authorization documentation up to date (there was some variation that doesn't actually exist)
* Added section about wildcard scopes to clarify that they aren't supported

### DIFF
--- a/content/dstu2.md
+++ b/content/dstu2.md
@@ -42,7 +42,7 @@ Transfer-Encoding: chunked
 
 Blank fields are omitted.
 
-All timestamps are returned in [FHIR<sup>®</sup> standard date/dateTime formats](http://www.hl7.org/implement/standards/fhir/datatypes.html#date).
+All timestamps are returned in [FHIR<sup>®</sup> standard date/dateTime formats](http://www.hl7.org/implement/standards/fhir/dstu2/datatypes.html#date).
 
 ### Media Types
 
@@ -60,13 +60,13 @@ application is a SMART application, the [service root url](http://hl7.org/fhir/d
 launch time. For standalone applications, the URL can be requested (or configured) when the application is set up to run
  against a specific tenant. FHIR calls will be made against URLs of the following format:
 
-`:serviceRootURL/:resource\[?:parameters]`
+`:serviceRootURL/:resource[?:parameters]`
 
 ### Open Sandbox
 
 The open sandbox instance allows developers to experiment with the service without requiring
 authentication. We recommend using this endpoint for initial proof of concepts and integration. The service root URL for
-this instance is: `https://fhir-open.sandboxcernerpowerchart.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/:resource?:parameters`
+this instance is: `https://fhir-open.sandboxcernerpowerchart.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/:resource[?:parameters]`
 
 Note: The open endpoint exposes read-only resources. No writes are available in sandbox without using authentication.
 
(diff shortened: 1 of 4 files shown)


content/may2015.md
```diff
@@ -40,7 +40,7 @@ Transfer-Encoding: chunked
 
 Blank fields are omitted.
 
-All timestamps are returned in [FHIR<sup>®</sup> standard date/dateTime format](http://www.hl7.org/implement/standards/fhir/datatypes.html#date).
+All timestamps are returned in [FHIR<sup>®</sup> standard date/dateTime formats](http://www.hl7.org/implement/standards/fhir/datatypes.html#date).
```
dennispatterson: This link should probably be http://www.hl7.org/implement/standards/fhir/2015may/datatypes.html#date

content/dstu2.md
```diff
@@ -53,24 +53,35 @@ Cerner supports the FHIR<sup>®</sup> standard defined media types for JSON cont
...
+## Service Root URL
 
-The URL takes the form `fhir.sandboxcernerpowerchart.com/dstu2/:ehr_source_id/:resource`.
+URLs for the FHIR server vary by the tenant (datasource or client) being accessed, as well as other factors. If the 
+application is a SMART application, the [service root url](http://hl7.org/fhir/dstu2/http.html#general) is provided at 
+launch time. For standalone applications, the URL can be requested (or configured) when the application is set up to run
+ against a specific tenant. FHIR calls will be made against URLs of the following format:
 
-### Open Endpoint
+`:serviceRootURL/:resource\[?:parameters]`
```
dennispatterson: Does that '\' after the resource need to be there?

content/may2015.md
```diff
@@ -78,8 +88,8 @@ parameter:
 $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcernerpowerchart.com/may2015/d075cf8b-3261-481d-97e5-ba6c48d3b41f/MedicationPrescription?patient=2744010&status=active"
 </pre>
 
-In this example, the 'd075cf8b-3261-481d-97e5-ba6c48d3b41f' value is provided for the `:ehr_source_id` parameter in the path
-while `patient` and `status` are passed in the query string.
+In this example, MedicationOrder is the FHIR<sup>®</sup> standard resource being accessed, while `patient` and `status`
```
dennispatterson: MedicationPrescription

content/dstu2.md
daliboz: It did before I put it in quotes (at least, to make it not report issues or highlight as a link in the ide). Removed :)

content/may2015.md
daliboz: fixed

content/may2015.md
daliboz: fixed both may and dstu2

daliboz:
Ok, the updates are finally pushed...

dennispatterson:
+1

content/may2015/authorization.md
```diff
@@ -93,54 +94,111 @@ The [bearer access token][7] returned from the authorization server is what you
...
 * refresh_token
 
 ```
 grant_type=refresh_token&refresh_token={REFRESH_TOKEN}
 ```
 
-If successful, the authorization server will return a **200 OK** response with a content-type of [**application/json**][3] similar to the example below. 
+If successful, the authorization server will return a **200 OK** response with a content-type of [**application/json**][3] similar to the non-normative example below.
 
 ```
```
MaxPhilips: We could change the code fencing to \<pre class="body-response">\<code class="language-javascript"> and \</code>\</pre> to prevent these token examples from displaying on a single massive line.

Also noticed that 'non-normative' was added here but not in DSTU2

daliboz: Since it's the id_token and access_token that are sliding off, I think this is ok. There's no real reason for them to need to see the actual tokens as much as to see the example response. I would think wrapping that base64 section may make it harder to read/understand the examples.

Fixed the non-normative

daliboz: Actually, I'm looking at the wrong examples :)

daliboz:
Ok, updates made

MaxPhilips:
+1

PR #76, opened 2016-09-06
Update DSTU2 Medication Statement Search documentation.

jfrhong:


lib/resources/dstu2/medication_statement_contained_practitioner.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+  - name: role
+    required: 'No'
+    cardinality: 0..1
+    type: CodeableConcept
+    description: The roles which this practitioner is authorized to perform for the organization.
+    binding:
+      description: The roles which this practitioner is authorized to perform for the organization.
+      terminology:
+          - display: Practitioner Role
```
daliboz: The requirements page states this is SNOMED-CT: https://wiki.ucern.com/display/Open/DSTU+2+%28Final%29+MedicationStatement+Resource#DSTU2(Final)MedicationStatementResource-PractitionerResource

And *not* the practioner-role system listed for the example binding.

daliboz:
+1 other than system issue.

lib/resources/dstu2/medication_statement_contained_practitioner.yaml
jfrhong: I think that was mistakenly copied from the AllergyIntolerance resource. Does Argonaut require us to use SNOMED? We told the mapping service to return the practitioner role codes and not SNOMED.

davischristi:
👍

PR #77, opened 2016-09-14
Update DSTU2 documentation for Conformance with new authorize token endpoints.

sravan559:


lib/dstu2_resources.rb
```diff
@@ -1300,22 +1373,22 @@ module Resources
                 {
                   "name": "patient",
                   "type": "reference",
-                  "documentation": "The identity of a patient to list dispenses for. It is a required field"
+                  "documentation": "The identity of a patient to list medication orders for. It is a required field"
```
joykoduru: Nit: End line with a period for consistency and I think that is the standard.

MaxPhilips: This data is just copied over from our conformance statement, so we should add the period in our actual server before adding it here.  No period is fine given our current conformance

sravan559: @MaxPhilips  👍

MaxPhilips:
+1

jfrhong:
+1

PR #83, opened 2016-11-03
Add initial Device resource  documentation.

davischristi:
Adding initial Device resource documentation.

### DIFF
--- a/lib/resources/dstu2/device.yaml
+++ b/lib/resources/dstu2/device.yaml
@@ -1,0 +1,141 @@
+---
+name: Device
+field_name_base_url: http://hl7.org/fhir/DSTU2/device-definitions.html#Device
+fields:
+- name: identifier
+  required: 'No'
+  cardinality: 0..*
+  type: Identifier
+  description: Identifier(s) from manufacturer, owner and others.
+  example: |
+    {
+      "identifier": [
+        {
+          "type": {
+            "coding": [
+              {
+                "system": "http://hl7.org/fhir/identifier-type",
+                "code": "SNO",
+                "display": "Serial Number"
+              }
+            ],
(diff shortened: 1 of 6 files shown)


lib/resources/dstu2/device.yaml
```diff
@@ -0,0 +1,141 @@
...
+            ],
+          "text": "Serial Number"
+        },
+        "value": "SERIALNBRDA88ABA9-9FAD-4A03-9E9A-B52299AE9AE4"
+        }
+      ]
+    }
+  children:
+
+  - name: type
```
daliboz: This isn't linking to anything right now - I don't think this is intended to be a child of identifier, right?

lib/resources/dstu2/device.yaml
```diff
@@ -0,0 +1,141 @@
...
+    {
+      "type": {
+        "text": "Cochlear Implant Right Ear"
+      }
+    }
+  binding:
+    description: The type of device.
+    note: Currently not bound to any terminology and only 'Device.type.text' will be returned as of now.
+
+- name: manufacturer
```
daliboz: Is this or below actually being used anywhere?

daliboz:
Should we add a note about not supporting the status modifier element?

lib/resources/dstu2/device.yaml
cerbs: Nope. They aren't used. I'll yank 'em

cerbs:
Added status modifier note

cerbs:
👍  After updates

cerbs:
Rebasing onto master

cerbs:
Merged and squashed on 224dc4c

PR #84, opened 2016-11-09
Contract - Documentation

cerbs:
Documentation for the Contract resource

- Also, the dstu_resources.rb file was becoming huge will all of the example JSON content into, so I broke it separate files with one for each resource. All the content is the same, just moved.


lib/resources/dstu2/contract.yaml
```diff
@@ -0,0 +1,56 @@
...
+  type: CodeableConcept
+  binding:
+    description: List of overall contract codes.
+    terminology:
+    - display: SNOMED CT
+      system: http://snomed.info/sct
+      values:
+      - 371537001 - Consent report (record artifact)
+
+- name: subtype
```
daliboz: The link being generated isn't valid, and ends up linking to the top of the Contract page

daliboz:
Need to add note to contract that it's currently only supported for record sharing type agreements? I know the bindings make this a bit obvious, but would be nice to have that called out. Our "template" supports an "overview" section now - this would be a good place for that as well

daliboz:
We probably need to set the search.mode on the entry to differentiate between matched and included resources: http://hl7.org/fhir/dstu2/bundle-definitions.html#Bundle.entry.search.mode, assuming examples are fully accurate, I don't see it. (enhancement)

daliboz:
Otherwise +1

cerbs:
Logged an issue to track on code change necessary for setting search.mode

content/dstu2/contract.md
```diff
@@ -0,0 +1,66 @@
...
+
+    GET /Contract?:parameters
+
+### Parameters
+
+ Name                 | Required?                                                    | Type          | Description
+----------------------|--------------------------------------------------------------|---------------|------------------------------------------------------------------------------------------------------------------------------
+ `_id`                | No, if populated all other parameters are ignored            | [`token`]     | The logical resource id associated with the resource.
+ `actor`              | Yes, or `_id`                                                | [`reference`] | The actor (RelatedPerson) of the Contract. Example: `actor=RelatedPerson/1234` or `actor:RelatedPerson=1234`
+ `actor.identifier`   | Yes, or `_id`                                                | [`token`]     | The identifier for the actor of the Contract. Example `actor.identifier=TODO`
```
MaxPhilips: TODO?

content/dstu2/contract.md
```diff
@@ -0,0 +1,66 @@
...
+ `patient.identifier` | Yes, or `_id`                                                | [`token`]     | The identifier for the patient subject of the Contract. Example `actor.identifier=TODO`
```
MaxPhilips: switch to patient.identifier and address TODO

lib/resources/example_json/dstu2_examples_allergy_intolerance.rb
```diff
@@ -0,0 +1,210 @@
...
+                {
+                    "system": "http://www.nlm.nih.gov/research/umls/rxnorm",
+                    "code": "3498"
+                }
+            ]
+        }
+    }
+    
+  end
+end
```
MaxPhilips: Some of the broken out examples need newlines at the ends of their files

content/dstu2/contract.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Either the `_id` or one of `actor`, `actor.identifier`, `patient.identifier`, or `subject` parameters must be provided.
+- The `actor`, `actor.identifier`, `patient.identifier`, or `subject` parameters may be provided exactly once and may have only a single value.
+- The `_include` parameter may be provided once for each resource to be included. Example: `_include=Contract:actor&_include=Contract:patient`
+- Additional OAuth2 scopes are required when the `_include` parameter is provided. To include with `Contract:actor` or `Contract:patient` values, the OAuth2 token must include scopes for one or both RelatedPerson.read and Patient.read, respectively, depending on the combination of resources to be included.
+
+### Response
+
+Search for all Contracts subject to a specific patient
+<%= headers 200, GET: '[...]/Contract?subject:Patient=14067775' %>
```
MaxPhilips: Fairly nitty but would it make more sense to have the GET line be first and the status second?

lib/resources/example_json/dstu2_examples_contract.rb
```diff
@@ -0,0 +1,613 @@
...
+          }
+        }
+      ]
+    }
+
+    DSTU2_CONTRACT_BUNDLE_INCLUDE_ACTOR ||= {
+      "resourceType": "Bundle",
+        "id": "bae9d7ad-79e4-4c85-9899-ef93f85b2dd5",
+        "type": "searchset",
+        "total": 2,
```
MaxPhilips: Is the total still 2 here?

content/dstu2/contract.md
MaxPhilips: No longer applicable

content/dstu2/contract.md
MaxPhilips: No longer applicable

MaxPhilips:
+1 after comments

lib/resources/example_json/dstu2_examples_contract.rb
cerbs: Yes. The total represents the number of search results and not the number of entries. There are 2 Contracts returned

lib/resources/example_json/dstu2_examples_allergy_intolerance.rb
cerbs: done

content/dstu2/contract.md
cerbs: Leaving as-is for now. This is reusing existing code and it's not quite as simple as just flipping the order. We can revisit in the future if necessary

cerbs:
Merged and squashed on a0736f2

PR #87, opened 2016-11-09
Update to document reference create description.

torresdrew:
Added more information to eh create description to explain that the document being created must be valid XHTML.

### DIFF
--- a/content/dstu2/document-reference.md
+++ b/content/dstu2/document-reference.md
@@ -13,7 +13,7 @@ title: DocumentReference | DSTU 2 API
 
 ## Create
 
-Create new documents. Currently only XHTML formatted documents are supported. Only documents that need to be associated to the patient's chart should be written. Please validate your document is valid XHTML using any available validator like the one linked [here]. 
+Create new documents. Currently limited to unstructured clinical notes or documentation. For example, a document with display formatting or styling can be written, but a CCD cannot. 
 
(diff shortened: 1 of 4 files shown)


content/dstu2/document-reference.md
```diff
@@ -13,7 +13,7 @@ title: DocumentReference | DSTU 2 API
 
 ## Create
 
-Create new documents. Only documents that would be classified as notes are currently supported.
+Create new documents. Currently only XHTML formatted documents are supported. Only documents that need to be associated to the patient's chart should be written. Please validate your document is valid XHTML using any available validator like the one linked [here].
```
daliboz: Maybe add this as a "note" similar with what we've done elsewhere?

We should also add it to the documentation for content.attachment.data - it's base64 encoded XHTML.

daliboz: Do we also want to update the May documentation?

torresdrew: Probably don't need to update May. Want to deprecate and don't think anyone is using writes on May. I will update with a note rather than changing the text.

daliboz: Why remove the comment about notes? This is currently a limitation of the service - it draws on the freetext note in the underlying code, I believe.

cerbs:
+1

torresdrew:
+1

PR #88, opened 2016-11-10
Update auth page to point to code console for application self regist…

mjhenkes:
…ration.

A bunch of spacing got changed. My changes are only around re-wording registration and changing the number 10 link to https://code.cerner.com/developer/smart-on-fhir/.

I updated the may 2015 page as well though it doesn't seem to be accessible anymore.

This shouldn't be merged until after the November 10th code console deployment.

### DIFF
--- a/content/dstu2/authorization.md
+++ b/content/dstu2/authorization.md
@@ -9,15 +9,9 @@ title: Authorization | DSTU 2 API
 The Cerner Authorization Server currently supports [OAuth 2.0][1] [SMART on FHIR<sup>®</sup>][4] launch workflows. As a client, it will require interaction between the client (your client application), the user, the authorization server, a SMART Launch server and a resource server that implements the FHIR<sup>®</sup> standard.
 
 ### Registration ###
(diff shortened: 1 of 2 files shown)


content/dstu2/authorization.md
```diff
@@ -9,15 +9,9 @@ title: Authorization | DSTU 2 API
 The Cerner Authorization Server currently supports [OAuth 2.0][1] [SMART on FHIR<sup>®</sup>][4] launch workflows. As a client, it will require interaction between the client (your client application), the user, the authorization server, a SMART Launch server and a resource server that implements the FHIR<sup>®</sup> standard.
 
 ### Registration ###
-In order for your client application to utilize any protected resources, your client application must first register. To do this please [contact us][10] with the following fields, all fields are **required**.
+In order for your client application to utilize any protected resources, your client application must first register using our [code console][10].
```
daliboz: Is there anything they should know about creating a Cerner Care account? I'm going to guess that most won't already have one. I know there's a big blue button, but it may be good to call that out a bit along with any expectations they would need to have (is that immediate?).

daliboz:
+1

PR #94, opened 2016-11-16
Updated invalid urls and ids with blank results to working ones on mu…

joykoduru:
…ltiple resources.

### DIFF
--- a/content/dstu2/contract.md
+++ b/content/dstu2/contract.md
@@ -42,11 +42,11 @@ Notes:
 ### Response
 
 Search for all Contracts subject to a specific patient
-<%= headers 200, GET: '[...]/Contract?subject:Patient=14067775' %>
+<%= headers 200, GET: '[...]/Contract?subject:Patient=4598008' %>
 <%= json(:dstu2_contract_bundle) %>
 
 Search for all Contracts and include referenced RelatedPerson entries in the response.
-<%= headers 200, GET: '[...]/Contract?subject:Patient=14067775&_include=Contract:actor' %>
+<%= headers 200, GET: '[...]/Contract?subject:Patient=4598008&_include=Contract:actor' %>
 <%= json(:dstu2_contract_bundle_include_actor) %>
 
 ## Retrieve by id
--- a/content/dstu2/person.md
+++ b/content/dstu2/person.md
@@ -25,12 +25,12 @@ Search for Persons that meet supplied query parameters:
  `identifier` | Yes, or `_id`                                     | [`token`]  | The person identifier.  Example: `urn:oid:2.16.840.1.113883.3.13.6|01022228`
 
 Notes:
- 
+
 - `identifier` value must include both a system and a code. Example: `identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:...:PI98N2FK5TN`
 
 ### Response
 
-<%= headers 200, GET: '[...]/Person?identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:IDENTITY-FEDERATION:REALM:AC193C1A-9763-45E1-9FA2-C1C3AA4BA16D:PRINCIPAL:R53HZ5ZL8ZX' %>
+<%= headers 200, GET: '[...]/Person?identifier=urn:oid:2.16.840.1.113883.3.13.6|urn:cerner:identity-federation:realm:687f29dd-69dd-4de5-acb1-fd8a2241ef3a:principal:uH3IW35N5RE' %>
 <%= json(:dstu2_person_bundle) %>
 
 ## Retrieve by id
@@ -41,7 +41,7 @@ List an individual Person by its id:
 
 ### Response
 
-<%= headers 200, GET: '[...]/Person/14149269' %>
+<%= headers 200, GET: '[...]/Person/4342009' %>
 <%= json(:dstu2_person_entry) %>
 
 [`token`]: http://hl7.org/fhir/DSTU2/search.html#token
--- a/content/dstu2/related-person.md
+++ b/content/dstu2/related-person.md
@@ -31,18 +31,18 @@ Notes:
 
 ### Response
 
-<%= headers 200, GET: '[...]/RelatedPerson?identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:IDENTITY-FEDERATION:REALM:AC193C1A-9763-45E1-9FA2-C1C3AA4BA16D:PRINCIPAL:PI98N2FK5TN' %>
+<%= headers 200, GET: '[...]/RelatedPerson?identifier=urn:oid:2.16.840.1.113883.3.13.6|urn:cerner:identity-federation:realm:687f29dd-69dd-4de5-acb1-fd8a2241ef3a:principal:EC4Ax54P8GI' %>
 <%= json(:dstu2_relatedperson_bundle) %>
 
 ## Retrieve by id
 
-List an individual RelatedPersons by its id:
+List an individual RelatedPerson by its id:
 
     GET /RelatedPerson/:id
 
 ### Response
 
-<%= headers 200, GET: '[...]/RelatedPerson/11325025' %>
+<%= headers 200, GET: '[...]/RelatedPerson/5796399' %>
 <%= json(:dstu2_relatedperson_entry) %>
 
 [`reference`]: http://hl7.org/fhir/DSTU2/search.html#reference
--- a/lib/resources/example_json/dstu2_examples_appointment.rb
+++ b/lib/resources/example_json/dstu2_examples_appointment.rb
@@ -1,50 +1,49 @@
 module Cerner
   module Resources
-   
+
     DSTU2_APPOINTMENT_BUNDLE ||= {
         "resourceType": "Bundle",
-        "id": "07f58859-b408-419f-b2ec-c0e22e3a179a",
+        "id": "67284a19-6a27-4638-9f04-9bff71324a58",
         "type": "searchset",
-        "total": 3,
+        "total": 1,
         "link": [
           {
             "relation": "self",
-            "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Appointment?date=2014&practitioner=18220111&status=arrived&_count=50"
+            "url": "https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Appointment?date=2007&practitioner=709932&_count=50"
           }
         ],
         "entry": [
           {
-            "fullUrl": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Appointment/20977581",
+            "fullUrl": "https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Appointment/759729",
             "resource": {
               "resourceType": "Appointment",
-              "id": "20977581",
+              "id": "759729",
               "meta": {
-                "versionId": "10",
-                "lastUpdated": "2014-05-22T10:04:29.000-05:00"
+                "versionId": "2",
+                "lastUpdated": "2007-02-28T20:04:18.000-06:00"
               },
               "text": {
                 "status": "generated",
-                "div": "&lt;div&gt;&lt;p&gt;&lt;b&gt;Appointment&lt;/b&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Description&lt;/b&gt;: Same Day&lt;/p&gt;&lt;p&gt;&lt;b&gt;Type&lt;/b&gt;: Same Day&lt;/p&gt;&lt;p&gt;&lt;b&gt;Start&lt;/b&gt;: 2014-05-22T09:53:00.000-05:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;End&lt;/b&gt;: 2014-05-22T10:53:00.000-05:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;Minutes Duration&lt;/b&gt;: 60&lt;/p&gt;&lt;p&gt;&lt;b&gt;Status&lt;/b&gt;: Arrived&lt;/p&gt;&lt;p&gt;&lt;b&gt;Reason&lt;/b&gt;: Headache&lt;/p&gt;&lt;p&gt;&lt;b&gt;Location&lt;/b&gt;: Baseline East&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Patient&lt;/b&gt;: Song, River&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Practitioner&lt;/b&gt;: Doctor, The&lt;/p&gt;&lt;/div&gt;"
+                "div": "&lt;div&gt;&lt;p&gt;&lt;b&gt;Appointment&lt;/b&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Description&lt;/b&gt;: CV Echo&lt;/p&gt;&lt;p&gt;&lt;b&gt;Type&lt;/b&gt;: CV Echo&lt;/p&gt;&lt;p&gt;&lt;b&gt;Start&lt;/b&gt;: 2007-02-28T08:00:00.000-06:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;End&lt;/b&gt;: 2007-02-28T08:15:00.000-06:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;Minutes Duration&lt;/b&gt;: 15&lt;/p&gt;&lt;p&gt;&lt;b&gt;Reason&lt;/b&gt;: Chest Pain&lt;/p&gt;&lt;p&gt;&lt;b&gt;Location&lt;/b&gt;: BE Cardiology&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Patient&lt;/b&gt;: Paddack, Dustin Lee&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Practitioner&lt;/b&gt;: Harris, Sarah&lt;/p&gt;&lt;p&gt;&lt;b&gt;Primary&lt;/b&gt;: true&lt;/p&gt;&lt;/div&gt;"
               },
-              "status": "arrived",
               "type": {
                 "coding": [
                   {
                     "system": "http://snomed.info/sct",
-                    "code": "394581000",
-                    "display": "Community medicine",
+                    "code": "394579002",
+                    "display": "Cardiology",
                     "userSelected": false
                   }
                 ],
-                "text": "Same Day"
+                "text": "CV Echo"
               },
               "reason": {
-                "text": "Headache"
+                "text": "Chest Pain"
               },
-              "description": "Same Day",
-              "start": "2014-05-22T09:53:00.000-05:00",
-              "end": "2014-05-22T10:53:00.000-05:00",
-              "minutesDuration": 60,
+              "description": "CV Echo",
+              "start": "2007-02-28T08:00:00.000-06:00",
+              "end": "2007-02-28T08:15:00.000-06:00",
+              "minutesDuration": 15,
               "participant": [
                 {
                   "type": [
@@ -63,8 +62,8 @@ module Cerner
                     }
                   ],
                   "actor": {
-                    "reference": "Practitioner/18220111",
-                    "display": "Doctor, The"
+                    "reference": "Practitioner/709932",
+                    "display": "Harris, Sarah"
                   },
                   "required": "required",
                   "status": "accepted"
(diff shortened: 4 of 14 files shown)


content/dstu2/related-person.md
```diff
@@ -42,7 +42,7 @@ List an individual RelatedPersons by its id:
 
 ### Response
 
-<%= headers 200, GET: '[...]/RelatedPerson?identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:IDENTITY-FEDERATION:REALM:AC193C1A-9763-45E1-9FA2-C1C3AA4BA16D:PRINCIPAL:PI98N2FK5TN' %>
+<%= headers 200, GET: '[...]/RelatedPerson?identifier=urn:oid:2.16.840.1.113883.3.13.6|NEW-FEDPRINCIPAL-ALIAS-001-MAX-VALIDATION' %>
```
MaxPhilips: This is an alias I created during RelatedPerson validation - can you find a different one?

content/dstu2/contract.md
```diff
@@ -42,11 +42,11 @@ Notes:
 ### Response
 
 Search for all Contracts subject to a specific patient
-<%= headers 200, GET: '[...]/Contract?subject:Patient=14067775' %>
+<%= headers 200, GET: '[...]/Contract?subject:Patient=4598008' %>
```
MaxPhilips: Is there a corresponding Contract example JSON update?

MaxPhilips: Also looks like the existing Contract JSON example is using FHIRPLAY, is that necessary @cerbs ?

cerbs: The g7 Resources were validated and stood up in FHIRPLAY. I don't know if they will work in SNDENG at all and there has not been build been done in that domain.

Additionally, I question the need for these changes altogether. Historically we have specifically avoided using valid data/ids in the responses because they could be linked to actual people with actual details. If we just "fix" the URLs to work then the results returned when someone calls the service will not match the JSON examples and start giving out data that don't want to give out.

I feel like the correct way to address this is to have custom build data configured specifically for these queries. Of course, good luck policing that when it comes to writes (see: DocumentReference)

daliboz: We usually use valid ids and urls so that someone could call them and see actual data. However, in the examples, we just make sure to adjust to obviously fake people to ensure someone doesn't Google a similar name (and find our indexed site) and think they've somehow found actual data (the data in the sanbox isn't real - but it sometimes looks real).

daliboz:
We need to update the appointment example as well - it needs to reference something in 0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca that has data

MaxPhilips:
As part of this pull request, can you also update:
* content/dstu2/appointment.md L40 "create a define" -> "define"
* content/dstu2/related-person.md L39 "individual RelatedPersons" -> "individual RelatedPerson"

content/dstu2/related-person.md
```diff
@@ -31,18 +31,18 @@ Notes:
...
 ### Response
 
-<%= headers 200, GET: '[...]/RelatedPerson/11325025' %>
+<%= headers 200, GET: '[...]/RelatedPerson/7470218' %>
 <%= json(:dstu2_relatedperson_entry) %>
 
 ## Retrieve by id
 
-List an individual RelatedPersons by its id:
+List an individual RelatedPersos by its id:
```
joykoduru: My bad. Fixed

content/dstu2/contract.md
MaxPhilips: The changed ids here still aren't in sync with the ids currently in the contract JSON examples

lib/resources/example_json/dstu2_examples_appointment.rb
```diff
@@ -1,50 +1,49 @@
...
-                "div": "&lt;div&gt;&lt;p&gt;&lt;b&gt;Appointment&lt;/b&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Description&lt;/b&gt;: Same Day&lt;/p&gt;&lt;p&gt;&lt;b&gt;Type&lt;/b&gt;: Same Day&lt;/p&gt;&lt;p&gt;&lt;b&gt;Start&lt;/b&gt;: 2014-05-22T09:53:00.000-05:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;End&lt;/b&gt;: 2014-05-22T10:53:00.000-05:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;Minutes Duration&lt;/b&gt;: 60&lt;/p&gt;&lt;p&gt;&lt;b&gt;Status&lt;/b&gt;: Arrived&lt;/p&gt;&lt;p&gt;&lt;b&gt;Reason&lt;/b&gt;: Headache&lt;/p&gt;&lt;p&gt;&lt;b&gt;Location&lt;/b&gt;: Baseline East&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Patient&lt;/b&gt;: Song, River&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Practitioner&lt;/b&gt;: Doctor, The&lt;/p&gt;&lt;/div&gt;"
+                "div": "<div><p><b>Appointment</b></p><p><b>Description</b>: CV Echo</p><p><b>Type</b>: CV Echo</p><p><b>Start</b>: 2007-02-28T08:00:00.000-06:00</p><p><b>End</b>: 2007-02-28T08:15:00.000-06:00</p><p><b>Minutes Duration</b>: 15</p><p><b>Reason</b>: Chest Pain</p><p><b>Location</b>: BE Cardiology</p><p></p><p><b>Patient</b>: Paddack, Dustin Lee</p><p></p><p><b>Practitioner</b>: Harris, Sarah</p><p><b>Primary</b>: true</p></div>"
```
MaxPhilips: Keep the escaped angled brackets otherwise they are treated as real HTML

lib/resources/example_json/dstu2_examples_appointment.rb
```diff
@@ -276,34 +97,25 @@ module Resources
...
           "status": "generated",
-          "div": "&lt;div&gt;&lt;p&gt;&lt;b&gt;Appointment&lt;/b&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Description&lt;/b&gt;: Same Day&lt;/p&gt;&lt;p&gt;&lt;b&gt;Type&lt;/b&gt;: Same Day&lt;/p&gt;&lt;p&gt;&lt;b&gt;Start&lt;/b&gt;: 2014-05-22T09:53:00.000-05:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;End&lt;/b&gt;: 2014-05-22T10:53:00.000-05:00&lt;/p&gt;&lt;p&gt;&lt;b&gt;Minutes Duration&lt;/b&gt;: 60&lt;/p&gt;&lt;p&gt;&lt;b&gt;Status&lt;/b&gt;: Arrived&lt;/p&gt;&lt;p&gt;&lt;b&gt;Reason&lt;/b&gt;: Headache&lt;/p&gt;&lt;p&gt;&lt;b&gt;Location&lt;/b&gt;: Baseline East&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Patient&lt;/b&gt;: Song, River&lt;/p&gt;&lt;p&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Practitioner&lt;/b&gt;: Doctor, The&lt;/p&gt;&lt;/div&gt;"
+          "div": "<div><p><b>Appointment</b></p><p><b>Description</b>: Echo</p><p><b>Type</b>: Echo</p><p><b>Start</b>: 2006-09-29T15:00:00.000-05:00</p><p><b>End</b>: 2006-09-29T15:15:00.000-05:00</p><p><b>Minutes Duration</b>: 15</p><p><b>Reason</b>: test</p><p><b>Location</b>: BE Cardiology</p><p></p><p><b>Patient</b>: PETERSON, FAITH</p><p></p><p><b>Practitioner</b>: Harris, Sarah</p><p><b>Primary</b>: true</p></div>"
```
MaxPhilips: Same here, escape angled brackets

lib/resources/example_json/dstu2_examples_condition.rb
```diff
@@ -9,7 +9,7 @@ module Resources
       "link": [
         {
           "relation": "self",
-          "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Condition?patient=2744123"
+          "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Condition?patient=1316024"
```
MaxPhilips: This URL returns a lot more than 3 results now

lib/resources/example_json/dstu2_examples_contract.rb
```diff
@@ -3,7 +3,7 @@ module Resources
 
     DSTU2_CONTRACT_ENTRY ||= {
       "resourceType": "Contract",
-      "id": "11325025",
+      "id": "7470218",
```
MaxPhilips: Can you update the rest of the response to match this new id?

lib/resources/example_json/dstu2_examples_immunization.rb
```diff
@@ -10,7 +10,7 @@ module Resources
         "link": [
           {
             "relation": "self",
-            "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Immunization?patient=2100031"
+            "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Immunization?patient=1316024"
```
MaxPhilips: 1316024 has 6 immunizations not 4

lib/resources/example_json/dstu2_examples_immunization.rb
```diff
@@ -269,7 +269,7 @@ module Resources
     DSTU2_IMMUNIZATION_1 ||=
       {
         "resourceType": "Immunization",
-        "id": "M22335399",
+        "id": "M5251276",
```
MaxPhilips: Can you update the rest of the example to match this id?

lib/resources/example_json/dstu2_examples_medication_order.rb
```diff
@@ -10,7 +10,7 @@ module Resources
         "link": [
           {
             "relation": "self",
-            "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/MedicationOrder?patient=1083999&_count=25"
+            "url": "https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/MedicationOrder?patient=4342008&_count=25"
```
MaxPhilips: This patient has 2 medication orders

lib/resources/example_json/dstu2_examples_patient.rb
```diff
@@ -3,7 +3,7 @@ module Resources
 
     DSTU2_PATIENT_1 ||= {
       "resourceType": "Patient",
-      "id": "9827347",
+      "id": "4596007",
```
MaxPhilips: This id no longer matches its entries' fullUrl

Also, the self link for our patient search example is a search by name=Carr but our examples have Dr. Who names (not Carr names).  Probably want to have the example search by something other than name

lib/resources/example_json/dstu2_examples_person.rb
```diff
@@ -67,7 +67,7 @@ module Resources
       ],
       "entry": [
         {
-          "fullUrl": "https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Person/14149269",
+          "fullUrl": "https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Person/4342009",
```
MaxPhilips: This fullUrl no longer matches the person id below

lib/resources/example_json/dstu2_examples_relatedperson.rb
```diff
@@ -3,7 +3,7 @@ module Resources
 
     DSTU2_RELATEDPERSON_ENTRY ||= {
       "resourceType": "RelatedPerson",
-      "id": "11325025",
+      "id": "7470218",
```
MaxPhilips: Can you update the rest of the example to match this new id?

joykoduru:
Addressed all comments, rebased and squashed

content/dstu2/related-person.md
```diff
@@ -31,18 +31,18 @@ Notes:
 
 ### Response
 
-<%= headers 200, GET: '[...]/RelatedPerson?identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:IDENTITY-FEDERATION:REALM:AC193C1A-9763-45E1-9FA2-C1C3AA4BA16D:PRINCIPAL:PI98N2FK5TN' %>
+<%= headers 200, GET: '[...]/RelatedPerson?_id=5796348' %>
```
MaxPhilips: @cerbs would we want to keep this example as an identifier search or is it ok to switch to search by _id?

lib/resources/example_json/dstu2_examples_contract.rb
```diff
@@ -498,58 +463,51 @@ module Resources
                     "text": "Federated Person Principal"
                   },
                   "system": "urn:oid:2.16.840.1.113883.3.13.6",
-                  "value": "URN:CERNER:IDENTITY-FEDERATION:REALM:AC193C1A-9763-45E1-9FA2-C1C3AA4BA16D:PRINCIPAL:R53HZ5ZL8ZX",
+                  "value": "NEW-FEDPRINCIPAL-ALIAS-001-MAX-VALIDATION",
```
MaxPhilips: Oops this is still leftover from our deployment verification.

I just patched in a new value that isn't so fake, can you update to "URN:CERNER:IDENTITY-FEDERATION:REALM:6A2A9D0D-C19E-4B52-BB97-22732FDAA8D6:PRINCIPAL:A50B81F9AD"

lib/resources/example_json/dstu2_examples_immunization.rb
```diff
@@ -4,151 +4,74 @@ module Resources
...
+                "reference": "Patient/4478007",
+                "display": "SMART, FRED RICK"
               },
               "wasNotGiven": false,
               "reported": false,
               "performer": {
-                "reference": "Practitioner/7750958",
-                "display": "Dr. Who"
+                "reference": "Practitioner/4474007",
+                "display": "Pickering, Kathy"
```
MaxPhilips: Can you switch this name?

lib/resources/example_json/dstu2_examples_immunization.rb
```diff
@@ -160,180 +83,70 @@ module Resources
...
+          "reference": "Patient/1316024",
+          "display": "PETERS, TIMOTHY"
         },
         "wasNotGiven": false,
         "reported": false,
         "performer": {
-          "reference": "Practitioner/7750958",
-          "display": "Dr. Who"
+          "reference": "Practitioner/2270007",
+          "display": "Davis, Kim"
```
MaxPhilips: Can you switch this name and its entry in the text above?

lib/resources/example_json/dstu2_examples_medication_order.rb
```diff
@@ -434,148 +96,63 @@ module Resources
...
-                "reference": "Patient/1083999",
-                "display": "River Song"
+                "reference": "Patient/4342008",
+                "display": "Smart, Wilma"
               },
               "prescriber": {
-                "reference": "Practitioner/709938",
-                "display": "Dr. Who"
+                "reference": "Practitioner/1912007",
+                "display": "McCurdy, Michael"
```
MaxPhilips: Can you switch this name and its entry in the text above?

lib/resources/example_json/dstu2_examples_patient.rb
```diff
@@ -1,23 +1,17 @@
 module Cerner
   module Resources
 
-    DSTU2_PATIENT_1 ||= {
+    DSTU2_PATIENT_ENTRY ||= {
       "resourceType": "Patient",
-      "id": "9827347",
+      "id": "4596007",
```
MaxPhilips: When I search for this id in SNDENG I get the patient "SMARTAPP, GrowthChartBMI"

lib/resources/example_json/dstu2_examples_patient.rb
```diff
@@ -30,254 +24,158 @@ module Resources
...
+                  "display": "Married",
+                  "userSelected": false
+                }
+              ],
+              "text": "Married"
+            },
+            "careProvider": [
+              {
+                "reference": "Practitioner/1912007",
+                "display": "McCurdy, Michael"
```
MaxPhilips: Can you update this name?

MaxPhilips:
Double check that all names of real people are swapped out with fake ones (e.g. can leave Timothy Peters but switch practitioner names)

content/dstu2/related-person.md
cerbs: I did intentionally use the identifier search it was the more interesting search and I don't really see much benefit in giving 2 examples basically showing the same thing. So my preference would be for it to be searching by identifier, but it isn't _wrong_ to use _id.

If we don't want to use identifier for whatever reason then I would prefer patient over _id

MaxPhilips: Hm I thought that I patched away all the MAX-VALIDATION identifiers but I guess they hang around as old versions

I think it will be understood if we used an example with a total of 40 but only included one or two actual results, replacing the other 38 with one instance of "..." or "<SNIPPED>"

lib/resources/example_json/dstu2_examples_relatedperson.rb
```diff
@@ -245,10 +245,11 @@ module Resources
               }
             ],
             "period": {
-              "start": "2015-10-30T06:48:46.000Z"
+              "start": "2015-10-30T06:50:09.000Z"
             }
           }
-        }
+        },
+        <SNIPPED>
```
MaxPhilips: Looks like my suggestion for <SNIPPED> will not work here as I am getting a nanoc compile error.  How about an ellipsis?

lib/resources/example_json/dstu2_examples_relatedperson.rb
```diff
@@ -120,27 +120,27 @@ module Resources
...
             "resourceType": "RelatedPerson",
-            "id": "5796348",
+            "id": "5796399",
             "meta": {
               "versionId": "30"
             },
             "text": {
               "status": "generated",
-              "div": "&lt;div&gt;&lt;p&gt;&lt;b&gt;RelatedPerson&lt;/b&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Name&lt;/b&gt;: PETERS, TIMOTHY&lt;/p&gt;&lt;p&gt;&lt;b&gt;DOB&lt;/b&gt;: 1961-04-30&lt;/p&gt;&lt;p&gt;&lt;b&gt;Sex&lt;/b&gt;: Male&lt;/p&gt;&lt;/div&gt;"
+              "div": "<div><p><b>RelatedPerson</b></p><p><b>Name</b>: PETERS, TIMOTHY</p><p><b>DOB</b>: 1961-04-30</p><p><b>Sex</b>: Male</p></div>"
```
MaxPhilips: The \<'s and \>'s have to be written as &lt; and &gt;

lib/resources/example_json/dstu2_examples_relatedperson.rb
```diff
@@ -183,7 +183,7 @@ module Resources
               }
             ],
             "patient": {
-              "reference": "Patient/3838010",
+              "reference": "Patient/3838011",
               "display": "Peters, Jennifer"
```
MaxPhilips: So... I know that this is actually what is returned by the url in question here. But it is probably a very confusing example to give, since the patient's name in the description is TIMOTHY and here is Jennifer. Probably just want to change this to Timothy even though that's different than what people will see if they fire off the request themselves. Or find another patient to use as an example

lib/resources/example_json/dstu2_examples_immunization.rb
```diff
@@ -120,7 +120,7 @@ module Resources
         "reported": false,
         "performer": {
           "reference": "Practitioner/2270007",
-          "display": "Davis, Kim"
+          "display": "Fhir, Enginne"
```
MaxPhilips: Engine

content/dstu2/person.md
```diff
@@ -25,12 +25,12 @@ Search for Persons that meet supplied query parameters:
...
 
 Notes:
- 
+
 - `identifier` value must include both a system and a code. Example: `identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:...:PI98N2FK5TN`
 
 ### Response
 
-<%= headers 200, GET: '[...]/Person?identifier=urn:oid:2.16.840.1.113883.3.13.6|URN:CERNER:IDENTITY-FEDERATION:REALM:AC193C1A-9763-45E1-9FA2-C1C3AA4BA16D:PRINCIPAL:R53HZ5ZL8ZX' %>
+<%= headers 200, GET: '[...]/Person?identifier=urn:oid:2.16.840.1.113883.3.13.6|urn:cerner:identity-federation:realm:687f29dd-69dd-4de5-acb1-fd8a2241ef3a:principal:uH3IW35N5RE' %>
```
VogelKC: Aren't these stored in the system in all uppercase? The example should reflect that.
(There are several instances of these in these files)

VogelKC:
Is there a particular reason were changing some of these examples to patients that have less data in their responses? Seems like it makes more sense for the documented examples to be more filled out.

content/dstu2/person.md
cerbs: This query is not case sensitive and in actuality this type of identifier is from an external system and is in-fact mixed case.

MaxPhilips:
I suggested giving examples that are limited to a single entry rather than a list of 40 patients.  I picked the id at the top of the list instead of examining each entry so perhaps we would want to select a person for example that does have a lot of optional fields populated

content/dstu2/person.md
VogelKC: Got it.

VogelKC:
Yeah, I understand slimming down the examples to reduce noise. It'd still be good to use samples that have more data (if possible). I don't know that I'm super-opinionated that it has to be done right now with this PR. It'd be better to get working URLs out then perfect sample patients :)

VogelKC:
Looks like there's still some changes left from the review Max requested?

lib/resources/example_json/dstu2_examples_contract.rb
joykoduru: Fixed in this [commit](https://github.com/cerner/fhir.cerner.com/pull/94/files#diff-7d5282157e9d95c6569ac37dbc6aba6dR113)

lib/resources/example_json/dstu2_examples_immunization.rb
joykoduru: After updating: [here](https://github.com/cerner/fhir.cerner.com/pull/94/files#diff-ec5da8065cea9cb0bfbb0695a022ebe1R96)

lib/resources/example_json/dstu2_examples_medication_order.rb
joykoduru: Fixed: [here](https://github.com/cerner/fhir.cerner.com/pull/94/commits/77142ee816a0d4601ef0490521187236fe963303#diff-d2c45c48de1b501ecb16965e88dc0165R9)

content/dstu2/contract.md
joykoduru: Example added: [here](https://github.com/cerner/fhir.cerner.com/pull/94/commits/77142ee816a0d4601ef0490521187236fe963303#diff-7d5282157e9d95c6569ac37dbc6aba6dR84)

VogelKC:
+1

lib/resources/example_json/dstu2_examples_relatedperson.rb
```diff
@@ -174,158 +176,76 @@ module Resources
...
-                {
-                  "system": "http://hl7.org/fhir/v3/RoleCode",
-                  "code": "DAUC",
-                  "display": "daughter",
-                  "userSelected": false
-                }
-              ],
-              "text": "Daughter"
+              "reference": "Patient/3838011",
+              "display": "Peters, Timothy"
```
MaxPhilips: @joykoduru I just realized that I was interpreting this incorrectly when I asked you to change this in the PR.  I didn't realize that this was a RelatedPerson example, so the display here should remain as Jennifer whoever.

MaxPhilips: Can you work with whoever is next to merge a PR to get this updated?

joykoduru: Sure. Currently there is a PR out for Goals. I will ask George/Davis to make this change.

PR #105, opened 2017-01-26
Add DSTU2 documentation for CarePlan

MaxPhilips:
Do we need to do anything special for the category terminology binding since we are using the argonaut extension?  Should I just include an example for http://argonaut.hl7.org|assess-plan ?

PR #107, opened 2017-02-02
Add documentation for Procedure resource

cerbs:
- Adding documentation for Procedure resource
- Fixing div encoding issue I noticed MedicationOrder

### DIFF
--- a/content/dstu2/procedure.md
+++ b/content/dstu2/procedure.md
@@ -1,0 +1,62 @@
+---
+title: Procedure | DSTU 2 API
+---
+
+# Procedure
+
+* TOC
+{:toc}
+
+## Terminology Bindings
+
+<%= terminology_table(:procedure, :dstu2) %>
+
+## Search
+
+Search for Procedures that meet supplied query parameters:
+
+    GET /Procedure?:parameters
+
+_Implementation Notes_
+
+* The [Procedure.notPerformed] modifier element is not supported and will not be returned.
+
+### Parameters
+
+ Name         | Required?                             | Type          | Description
+--------------|---------------------------------------|---------------|------------------------------------------------------------------------------------
+ `_id`        | This, or one of `patient` or `subject`| [`token`]     | The logical resource id associated with the resource. Example: `_id=24110557`
+ `patient`    | This, or one of `_id` or `subject`    | [`reference`] | The patient subject of the Procedure. Example: `patient=1316024`
+ `subject`    | This, or one of `_id` or `patient`    | [`reference`] | The subject of the Procedure. Must represent a Patient resource. May use the `:Patient` modifier. Example: `subject=Patient/1316020` or `subject:Patient=1316020`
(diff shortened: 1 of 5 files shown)


MaxPhilips:
The MedOrder div fix is also included here: https://github.com/cerner/fhir.cerner.com/pull/105#issuecomment-277006616

Race ya

content/dstu2/procedure.md
```diff
@@ -0,0 +1,62 @@
...
+* The [Procedure.notPerformed] modifier element is not supported and will not be returned.
+
+### Parameters
+
+ Name         | Required?                             | Type          | Description
+--------------|---------------------------------------|---------------|------------------------------------------------------------------------------------
+ `_id`        | This, or one of `patient` or `subject`| [`token`]     | The logical resource id associated with the resource. Example: `_id=24110557`
+ `patient`    | This, or one of `_id` or `subject`    | [`reference`] | The patient subject of the Procedure. Example: `patient=1316024`
+ `subject`    | This, or one of `_id` or `patient`    | [`reference`] | The subject of the Procedure. Must represent a Patient resource. May use the `:Patient` modifier. Example: `subject=Patient/1316020` or `subject:Patient=1316020`
+ `date`       | No                                    | [`date`]      | The date/time when the Procedure was performed. Must use the `ge` and/or `le` prefixes. Example: `date=le2017-02-01T10:30:00Z`
```
MaxPhilips: Can you change this to an 'N' to match other pages?

cerbs: Will change to N, but should note that 1) they don't all follow that pattern, and 2) it would probably be better to change all the other ones from Y/N to Yes/No

lib/resources/dstu2/procedure.yaml
```diff
@@ -0,0 +1,23 @@
...
+    description: A code to identify a specific procedure
+    terminology:
+    - display: Procedure Codes (SNOMED CT)
+      system: http://snomed.info/sct
+    - display: CPT
+      system: http://www.ama-assn.org/go/cpt
+    - display: ICD-10-PCS
+      system: http://www.icd10data.com/icd10pcs
+
+- name: performer
```
MaxPhilips: This isn't getting used for the terminology binding table currently, do we need it?

MaxPhilips:
+1 after comments

content/dstu2/procedure.md
MaxPhilips: I'd prefer Yes/No as well, let's decide on that strategy going forward and update the existing pages

lib/resources/dstu2/procedure.yaml
cerbs: It was supposed to be. I missed the terminology tag (and updating the system)

davischristi:
👍

cerbs:
Merged with 9b76a80

PR #111, opened 2017-02-28
Proposed changes to registration text, corrected previous spelling er…

whitehatguy:
…ror of 'misuses'


daliboz:
We updated the wording a bit on the original PR.

PR #113, opened 2017-03-06
Add documentation for health concerns and retrieve by id functionalit…

davischristi:
…y for Condition resource.

Added documentation for:
- Retrieve by id functionality
- Support for ```health-concern``` category code
- Updated conformance to include ```_id``` search parameter.

### DIFF
--- a/content/dstu2/condition.md
+++ b/content/dstu2/condition.md
@@ -19,18 +19,36 @@ Search for Conditions that meet supplied query parameters:
 
 ### Parameters
 
- Name             | Required? | Type          | Description
-------------------|-----------|---------------|------------------------------------------------------------------
- `patient`        | Y         | [`reference`] | The patient who has the condition. Example: `12345`
- `category`       | N         | [`token`]     | The [category] of the condition. Example: `diagnosis`, `problem`
- `clinicalstatus` | N         | [`token`]     | The [clinical status] of the condition. Example: `resolved`
+ Name             | Required?                                                            | Type          | Description
+------------------|----------------------------------------------------------------------|---------------|----------------------------------------------------------------------------------------------------------------
+ `_id`            | This or `patient`, if populated all other parameters are not allowed | [`token`]     | The logical resource id associated with the resource. Example: `p12345`
+ `patient`        | This or `_id`                                                        | [`reference`] | The patient who has the condition. Example: `12345`
+ `category`       | N                                                                    | [`token`]     | The [category] of the condition. Example: `diagnosis`, `problem`, `health-concern`
+ `clinicalstatus` | N                                                                    | [`token`]     | The [clinical status] of the condition. Example: `resolved`
 
-Note: Currently `diagnosis` and `problem` category codes are supported. Code `diagnosis` is from the [condition-category] system and code `problem` is from the [argonaut extension-codes] system.
+Note: Currently `diagnosis`, `problem` and `health-concern` category codes are supported.
 
 ### Response
 
-<%= headers 200 %>
+<%= headers 200, {Functionality: 'Search by patient'} %>
 <%= json(:dstu2_condition_bundle) %>
+<%= headers 200, {Functionality: 'Search by id'} %>
+<%= json(:dstu2_condition_bundle_by_id) %>
+
+## Retrieve by id
+
+List an individual Condition by its id:
+
+    GET /Condition/:id
+
+### Response
+
+<%= headers 200, {Functionality: 'Retrieve diagnosis by id'} %>
+<%= json(:dstu2_condition_diagnosis_resource) %>
+<%= headers 200, {Functionality: 'Retrieve health concern by id'} %>
+<%= json(:dstu2_condition_health_concern_resource) %>
+<%= headers 200, {Functionality: 'Retrieve problem by id'} %>
+<%= json(:dstu2_condition_problem_resource) %>
 
 ## Create
 
@@ -44,6 +62,10 @@ To successfully POST a condition, the following headers must be provided. Condit
 
     Content-Type: application/json+fhir
     Authorization: <OAuth2 Bearer Token>
+
+_Implementation Notes_
+
+* Currently, `health-concern` category code is not supported for writing conditions.
 
 ### Body fields
 
@@ -125,6 +147,10 @@ To successfully PUT a condition, the following headers must be provided. Conditi
     Authorization: <OAuth2 Bearer Token>
     If-Match: W/"<Current condition id>"
 
+_Implementation Notes_
+
+* Currently, `health-concern` category code is not supported for updating conditions.
+
 ### Body fields
 
 <%= definition_table(:condition, :update, :dstu2) %>
@@ -166,5 +192,4 @@ Note: abatementDateTime is not supported for diagnosis
 [category]: http://hl7.org/fhir/DSTU2/valueset-condition-category.html
 [clinical status]: http://hl7.org/fhir/DSTU2/valueset-condition-clinical.html
 [condition-category]: http://hl7.org/fhir/condition-category
-[argonaut extension-codes]: http://argonaut.hl7.org/extension-codes
 [FHIR<sup>®</sup> Update]: http://hl7.org/fhir/DSTU2/http.html#update
--- a/lib/resources/dstu2/condition.yaml
+++ b/lib/resources/dstu2/condition.yaml
@@ -81,7 +81,7 @@ fields:
   required: 'Yes'
   cardinality: 1..1
   type: CodeableConcept
-  description: Identification of the condition, problem or diagnosis.
+  description: Identification of the condition, problem, diagnosis or health concern.
   example: |
     {
       "code": {
@@ -129,7 +129,7 @@ fields:
       "category": {
         "coding": [
           {
-            "system": "http://argonaut.hl7.org/extension-codes",
+            "system": "http://argonaut.hl7.org",
             "code": "problem",
             "display": "Problem"
           }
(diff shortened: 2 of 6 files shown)


content/dstu2/condition.md
```diff
@@ -166,5 +192,5 @@ Note: abatementDateTime is not supported for diagnosis
 [category]: http://hl7.org/fhir/DSTU2/valueset-condition-category.html
 [clinical status]: http://hl7.org/fhir/DSTU2/valueset-condition-clinical.html
 [condition-category]: http://hl7.org/fhir/condition-category
-[argonaut extension-codes]: http://argonaut.hl7.org/extension-codes
+[argonaut]: http://argonaut.hl7.org
```
MaxPhilips: This link doesn't go anywhere, is it still worth displaying it as a hyperlink in the note? I know the previous one didn't lead anywhere either but as a consumer if I see a hyperlink I would expect to be able to click it to read more.

lib/resources/dstu2/condition.yaml
```diff
@@ -129,22 +129,35 @@ fields:
...
+            "system": "http://argonaut.hl7.org",
             "code": "problem",
             "display": "Problem"
           }
         ],
         "text": "Problem"
       }
     }
+    OR
+    {
```
MaxPhilips: If the health-concern category is not supported for writes, should we be calling it out here?

lib/resources/example_json/dstu2_examples_condition.rb
```diff
@@ -1,11 +1,128 @@
...
+    DSTU2_CONDITION_HEALTH_CONCERN_RESOURCE = {
+      "resourceType": "Condition",
+      "id": "7965c636-f343-4e96-8904-3bc35e9fd9f9",
+      "meta": {
+        "versionId": "7965c636-f343-4e96-8904-3bc35e9fd9f9",
+        "lastUpdated": "2017-01-11T09:09:54.000Z"
+      },
+      "text": {
+        "status": "generated",
+        "div": "&lt;div&gt;&lt;p&gt;&lt;b&gt;Condition&lt;/b&gt;&lt;/p&gt;&lt;p&gt;&lt;b&gt;Patient&lt;/b&gt;: DND Pat, shams&lt;/p&gt;&lt;p&gt;&lt;b&gt;Health Concern&lt;/b&gt;: Decs&lt;/p&gt;&lt;p&gt;&lt;b&gt;Clinical Status&lt;/b&gt;: Active&lt;/p&gt;&lt;p&gt;&lt;b&gt;Verification Status&lt;/b&gt;: Confirmed&lt;/p&gt;&lt;p&gt;&lt;b&gt;Onset&lt;/b&gt;: 2017-01-11&lt;/p&gt;&lt;p&gt;&lt;b&gt;Resolved&lt;/b&gt;: false&lt;/p&gt;&lt;/div&gt;"
```
MaxPhilips: Different patient in the div compared to the patient field

content/dstu2/condition.md
davischristi: yeah point taken. But I think we should still link to the correct system that it relates to. Open for discussion if we want to hyperlink it or just keep it as a plain text.

lib/resources/dstu2/condition.yaml
davischristi: We have called it out as implementation note as a whole for the create and update functionality separately and also as a note on the binding as well on line 155.

lib/resources/example_json/dstu2_examples_condition.rb
davischristi: Fixed.

content/dstu2/condition.md
MaxPhilips: My opinion would be to avoid hyperlinking to something that doesn't load

MaxPhilips:
+1 after a resolution for the health-concern write/update example

content/dstu2/condition.md
cerbs: There's no expectation that systems are actually URLs. For instance the system for Patient.communication.language is just "urn:ietf:bcp:47".  We should probably be linking to the valueset definitions rather than the systems. If we need to call out the systems specifically it would probably be better for them to be unlinked free text.

Linking to nothing doesn't provide any benefit. It just looks broken.

lib/resources/dstu2/condition.yaml
```diff
@@ -129,22 +129,35 @@ fields:
...
+            "display": "Health Concern"
+          }
+        ],
+        "text": "Health Concern"
+      }
+    }
   binding:
     description: A category assigned to the condition.
-    note: Currently `diagnosis` and `problem` category codes are supported. Code `diagnosis` is bound to condition category codes and `problem` is bound to argonaut extenstion codes terminology system shown below.
+    note: Category codes `diagnosis`, `problem` and `health-concern` are currently supported for search and retrieve functionality. Code `health-concern` is currently not supported for conditions add and update functionality. Code `diagnosis` is bound to condition category codes where as codes `problem` and `health-concern` are bound to argonaut terminology system shown below.
```
cerbs: "where as" => "whereas"

content/dstu2/condition.md
```diff
@@ -19,18 +19,36 @@ Search for Conditions that meet supplied query parameters:
...
+ `_id`            | This or `patient`, if populated all other parameters are not allowed | [`token`]     | The logical resource id associated with the resource. Example: `p12345`
+ `patient`        | This or `_id`                                                        | [`reference`] | The patient who has the condition. Example: `12345`
+ `category`       | N                                                                    | [`token`]     | The [category] of the condition. Example: `diagnosis`, `problem`, `health-concern`
+ `clinicalstatus` | N                                                                    | [`token`]     | The [clinical status] of the condition. Example: `resolved`
 
-Note: Currently `diagnosis` and `problem` category codes are supported. Code `diagnosis` is from the [condition-category] system and code `problem` is from the [argonaut extension-codes] system.
+Note: Currently `diagnosis`, `problem` and `health-concern` category codes are supported. Code `diagnosis` is from the [condition-category] system and codes `problem` and `health-concern` are from the [argonaut] system.
```
cerbs: Looking at the generated site I think I would rather see \[argonaut] replaced with \`http://argonaut.hl7.org/`. Chrome doesn't even display the full system if you hover over the link. It strips of the http://.

cerbs: Also, Argonaut is a proper name and should be capitalized when used. Are there any licensing concerns with it? Do we need registration marks when we use it?

daliboz: This just feels overly confusing - they can look at the value sets if they want by clicking the links (they should be able to, at least). Good to call out supported categories. No need to mention which systems they're all from. No need to call out Argonaut as something separate.

content/dstu2/condition.md
```diff
@@ -125,6 +147,10 @@ To successfully PUT a condition, the following headers must be provided. Conditi
     Authorization: <OAuth2 Bearer Token>
     If-Match: W/"<Current condition id>"
 
+_Implementation Notes_
+
+* Currently, `health-concern` category code from the [argonaut] system is not supported for updating conditions.
```
daliboz: Just call out that health-concern is not supported for update. No need to call out the system. It's not ambiguous.

content/dstu2/condition.md
```diff
@@ -45,6 +63,10 @@ To successfully POST a condition, the following headers must be provided. Condit
     Content-Type: application/json+fhir
     Authorization: <OAuth2 Bearer Token>
 
+_Implementation Notes_
+
+* Currently, `health-concern` category code from the [argonaut] system is not supported for writing conditions.
```
daliboz: Same comment about not calling out the system

content/dstu2/condition.md
daliboz: +1 to linking to the value set definitions. I think this is what we've done when we link to SNOMED etc.

davischristi: Should we link it to http://argonautwiki.hl7.org/index.php?title=Problems_and_Health_Concerns#problems_and_health_concerns_code_extensions ?

content/dstu2/condition.md
davischristi: Fixed.

content/dstu2/condition.md
davischristi: Fixed.

content/dstu2/condition.md
davischristi: Fixed.

davischristi:
Addressed PR comments.

lib/resources/dstu2/condition.yaml
```diff
@@ -139,12 +139,12 @@ fields:
     }
   binding:
     description: A category assigned to the condition.
-    note: Currently `diagnosis` and `problem` category codes are supported. Code `diagnosis` is bound to condition category codes and `problem` is bound to argonaut extenstion codes terminology system shown below.
+    note: Category codes `diagnosis`, `problem` and `health-concern` are currently supported for search and retrieve functionality. Code `health-concern` is currently not supported for conditions add and update functionality.
```
cerbs: This note is unclear. Can drop the "currently"s and it's unclear what category codes are supported for create and update. Diagnosis and problem are explicitly included out for search and retrieve. Is the exclusion of health-concern from create and update implying that diagnosis and problem are supported?

lib/resources/dstu2/condition.yaml
```diff
@@ -139,12 +139,12 @@ fields:
...
     terminology:
     - display: Condition Category Codes
       system: http://hl7.org/fhir/condition-category
-    - display: Argonaut Extension Code
-      system: http://argonaut.hl7.org/extension-codes
+    - display: Argonaut Valueset Codes
+      system: http://argonautwiki.hl7.org/index.php?title=Problems_and_Health_Concerns#problems_and_health_concerns_code_extensions
```
cerbs: Is this the current link we should be using. Drew has always been pointing me to pages at http://build.fhir.org/ig/Healthedata1/Argo-DSTU2/ when referencing Argonaut implementation guides

lib/resources/dstu2/condition.yaml
daliboz: I don't think the definition of what action supports which category belongs here. That should be in the implementation notes of each action.

lib/resources/dstu2/condition.yaml
daliboz: How about... neither

ValueSet:
http://www.fhir.org/guides/argonaut/r2/ValueSet-condition-category.html

Official IG: http://www.fhir.org/guides/argonaut/r2/

lib/resources/dstu2/condition.yaml
davischristi: Updated the note.

lib/resources/dstu2/condition.yaml
davischristi: Updated to point to the correct valueset link.

davischristi:
Addressed PR comments.

MaxPhilips:
+1

cerbs:
+1

davischristi:
Squashed the commits. Rebased on master.

PR #115, opened 2017-03-07
Rework Navigation to separate api and api versions

daliboz:
Reworking the way we navigate our site, preparing for other FHIR implementations and shared auth content.

This looks daunting, but it is comprised of a large set of:
* Removals: May ballot documentation will be removed now that it's no longer supported
* Moves: We now have 2 levels of nav, and will group resources the way they're categorized by the FHIR spec. The folder structure reflects this. There are some link changes in some pages because of this.
* Redirects: The old locations will redirect to the new locations after a brief redirect page
* Additions: Each category now has an high level overview
* Cleanup: We had a lot of js we didn't use, or that didn't apply to the way we organized our site.


daliboz:
Should we redirect all May endpoints to a static page that states it's deprecated and to see the DSTU 2 documentation?

daliboz:
I need to move the general-clinical section to be alphabetical :)

lib/helpers/attributes_helper.rb
```diff
@@ -0,0 +1,21 @@
+module AttributesHelper
+  def add_version_attribute
+    @items.each do |item|
+      if item.identifier.to_s.include? '/dstu2'
+        item[:version] = 'dstu2'
+      elsif item.identifier.to_s.include? '/may2015'
```
dennispatterson: Do we need this anymore if all the may pages are deleted?

dennispatterson:
Does General-Clinical need to be alphabetical?  If we're starting to use FHIR's categories (which is good), any thoughts on using some of the order laid out here for listing the categories?: http://hl7.org/fhir/dstu2/resourcelist.html .  Or, looking toward STU3, there's this order: http://build.fhir.org/resourcelist.html

layouts/millennium_dstu2_sidebar.html
```diff
@@ -0,0 +1,98 @@
...
+          <li><a href="/millennium/dstu2/individuals/patient/">Patient</a></li>
+          <li><a href="/millennium/dstu2/individuals/related-person/">RelatedPerson</a></li>
+        </ul>
+      </li>
+      <li class="js-topic">
+        <h3><a href="#" class="js-expand-btn collapsed arrow-btn" data-proofer-ignore></a><a href="/millennium/dstu2/medications/">Medications</a>
+        </h3>
+        <ul class="js-guides">
+          <li><a href="/millennium/dstu2/medications/immunization/">Immunization</a></li>
+          <li><a href="/millennium/dstu2/medications/medication-statement/">MedicationStatement</a></li>
```
dennispatterson: nit: since you were the one to bring up alphabetical order, swap this with medication order. :)

lib/helpers/attributes_helper.rb
daliboz: No, I can remove the may check.

layouts/millennium_dstu2_sidebar.html
daliboz: yup :)

daliboz:
The downside of that page is that it's not completely accurate - the better classification is here: http://hl7.org/fhir/dstu2/resourceguide.html

I felt like alphabetical was a good compromise between "easy to find" and "really in depth knowledge of FHIR's classifications"

On that note - I left conformance, op outcome, (and structure definition, if we add it) on the overview section, though they would belong to a different group. Thoughts?

dennispatterson:
That's fine.  I'll concede the order. And, we can always extract the resources under overview if they grow in number.

On the topic of the redirect for May pages, that will be a good idea because Google Analytics show that those pages still get hit.

MaxPhilips:
It shouldn't matter when rendered, but some of the resource category pages have extra blank lines in their markdown (Entities and General Clinical)

daliboz:
rebased and updated per comments

dennispatterson:
👍

MaxPhilips:
The faq link from layouts/highlights.html needs to be updated from /faq/ to /millennium/faq/

MaxPhilips:
patient link on L320 of content/millennium/dstu2/dstu2.md needs to be updated to include /millennium/

MaxPhilips:
authorization link on L389 of content/millennium/dstu2/dstu2.md needs to be updated to drop /dstu2/

MaxPhilips:
discovery link on L917 of content/millennium/authorization.md needs to be updated to add /authorization/

MaxPhilips:
+1, those are the only links I saw that were still troublesome

PR #121, opened 2017-03-20
Add DSTU2 documentation about uniqueness of ids and identifiers

joykoduru:


daliboz:
I'm not sure I'm following the identifier example. In millennium, there are only a few identifiers that would have a uniqueness guarantee - and even then, with merge/unmerge etc, the thing pointed at (eg: patient) could change. 

Also, why use the _id param instead of the "self" url that each resource should reference? I don't think there's an expectation for a query to be unique or not unique - that identifies a bundle, not a specific resource.

dennispatterson:
Documenting some of our discussion yesterday:

The purpose of this is to document that the full url must be taken into consideration, not merely the resource + value.  So, the "not unique" examples would be something like "Patient/1316024", whereas the unique version would have the full url.

To Jenni's point, we aren't trying to make statements guaranteeing uniqueness of identifer searches, just emphasizing that identifiers also must be in the context of the full url.

Lastly, we didn't get to discuss this, but on the jira I mentioned having 4 examples.  We don't need multiple examples for each of those 4 cases.

content/dstu2.md
```diff
@@ -322,6 +322,28 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
 In this example, MedicationOrder is the FHIR<sup>®</sup> standard resource being accessed, while `patient` and `status` 
  are passed in the query string.
 
+### Resource Identity
+ 
+Please note that, in order to determine the uniqueness of ids and identifiers in Millennium EHR, complete URL must be considered. A complete URL comprises of [service root url](http://hl7.org/fhir/dstu2/http.html#general) and parameters (if any).
```
dennispatterson: url also includes the resource

content/dstu2.md
```diff
@@ -322,6 +322,28 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
...
+
+For example,
+
+To determine uniquesness of an id, consider:
+
+    https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Person?_id=Patient/1316024
```
dennispatterson: To Jenni's comment, this kind of search returns a Bundle and not a single resource.  You're also querying Person by Patient id

content/dstu2.md
```diff
@@ -322,6 +322,28 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
...
+
+instead of:
+
+    Patien/1316024
```
dennispatterson: Patien -> Patient

content/dstu2.md
```diff
@@ -322,6 +322,28 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
...
+
+To determine uniquesness of an identifier, consider:
```
dennispatterson: Take a look over Jenni's comment again about not guaranteeing uniqueness of identifiers.

content/millennium/dstu2.md
```diff
@@ -329,9 +329,29 @@ parameter:
...
+
+ For example, one must take into account the entire url and not simply the id or resource + id:
+
+     https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Person/1316024
+
+ In another context the id "Person/1316024" may identify another person entirely. In the following example a different resource may be returned because the context (service root url) has changed.
+
+     https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Person/1316024
+
+ Similarly when considering an identifier one must consider its context even when normally it might be considered unique (social security number etc.):
```
dennispatterson: While social security number is an example of an identifier, one would hope that would be unique across multiple urls.  MRN might be a better example, as this will be organization-specific, rather than potentially existing across multiple systems.  So, rather than discuss uniqueness, you might explain that while some identifiers may exist across systems, this is not guaranteed, and (as above), could result in identifying different patients.  Therefore, they should be used in the context of the full url.

content/millennium/dstu2.md
```diff
@@ -329,9 +329,29 @@ parameter:
...
+In this example, MedicationOrder is the FHIR<sup>®</sup> standard resource being accessed, while `patient` and `status`
  are passed in the query string.
 
+ ### Resource Identity
+
+ Please note that no ids or identifiers in Millennium EHR are intended to be used outside of the context of their complete URL. A complete URL comprises of [service root url](http://hl7.org/fhir/dstu2/http.html#general), the resource, and parameters (if any).
+
+ For example, one must take into account the entire url and not simply the id or resource + id:
+
+     https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Person/1316024
```
dennispatterson: To keep consistency across examples, let's use Patient for the by-id example, as well.

content/millennium/dstu2.md
```diff
@@ -332,6 +332,26 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
...
+
+ In another context the id "Patient/1316024" may identify another person entirely. In the following example a different resource may be returned because the context (service root url) has changed.
+
+     https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Patient/1316024
+
+ Similarly when considering an identifier one must consider it only in its full context. Even though some identifiers may exist across multiple systems (ex: MRN) it is not guaranteed that they will refer to the same resource.
+
+     https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Patient?identifier=urn:oid:1.1.1.1.1.1|10002700
+
+ For example, when using the above MRN in a different system, we are not guaranteed that the same patient resource is returned:
```
dennispatterson: Since this is a search that returns a Bundle and not just a resource, you could change it to "...same Patient resource is returned in the response Bundle:"

dennispatterson:
After the single change is made for the comment above, +1

content/millennium/dstu2.md
```diff
@@ -332,6 +332,26 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
 In this example, MedicationOrder is the FHIR<sup>®</sup> standard resource being accessed, while `patient` and `status` 
  are passed in the query string.
 
+ ### Resource Identity
+
+ Please note that no ids or identifiers in Millennium EHR are intended to be used outside of the context of their complete URL. A complete URL comprises of [service root url](http://hl7.org/fhir/dstu2/http.html#general), the resource, and parameters (if any).
```
MaxPhilips: grammar nit: 'comprises of' sounds odd to me, can we switch to either 'comprises the' or 'is comprised of the' and add 'the' before parameters?

content/millennium/dstu2.md
```diff
@@ -332,6 +332,26 @@ $ curl -i -H "Accept: application/json+fhir" "https://fhir-open.sandboxcerner.co
...
+
+ For example, one must take into account the entire url and not simply the id or resource + id:
+
+     https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Patient/1316024
+
+ In another context the id "Patient/1316024" may identify another person entirely. In the following example a different resource may be returned because the context (service root url) has changed.
+
+     https://fhir-open.sandboxcerner.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/Patient/1316024
```
MaxPhilips: Do we still want to provide this tenant key as an example? Soon consumers won't be able to send this request off

MaxPhilips:
+1 after comments

content/millennium/dstu2.md
dennispatterson: There's no other sandbox tenant we could use because FHIRPlay will soon be the only available public source of data.  I think the primary use-case for these examples is for illustration, not for execution.  If we are concerned about sndeng's tenant being used, the alternative would be to use completely fake tenants and just have abc123 and def456 or something.  Using fhirplay and sndeng at least makes one example work and uses a realistic style tenant for the other example.

MaxPhilips: Yeah I think this is alright, just wanted to bring it up since we have been updating other examples on the site away from sndeng

JFisk42:
Changes have been made. Squashing and merging.

PR #122, opened 2017-03-21
Apply template to Patient, add authorization types

daliboz:
Apply the common template to the Patient resource, and add the Authorization types supported per action

### DIFF
--- a/content/millennium/dstu2/individuals/patient.md
+++ b/content/millennium/dstu2/individuals/patient.md
@@ -77,10 +77,7 @@ Notes:
 
 ### Headers
 
-To successfully GET a Patient, the following headers must be provided. Authorization must be provided if accessing the closed endpoint.
-
-        Accept-Type: application/json+fhir
-        Authorization: <OAuth2 Bearer Token>
+<%= headers %>
 
 ### Example
 
@@ -90,12 +87,12 @@ To successfully GET a Patient, the following headers must be provided. Authoriza
 
 #### Response
 
-<%= headers 200 %>
+<%= headers status: 200 %>
 <%= json(:dstu2_patient_bundle) %>
 
 ### Errors
 
-The common [`errors`] may be returned.
+The common [errors] may be returned.
 
 ## Retrieve by id
 
@@ -113,19 +110,17 @@ _Implementation Notes_
 
 ### Headers
 
-To successfully GET a Patient, the following headers must be provided. Authorization must be provided if accessing the closed endpoint.
-
-        Accept-Type: application/json+fhir
-        Authorization: <OAuth2 Bearer Token>
+<%= headers %>
 
 ### Example
+
 #### Request
 
     GET https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Patient/4342008
 
 #### Response
 
-<%= headers 200 %>
+<%= headers status: 200 %>
 <%= json(:dstu2_patient_entry) %>
 
 ### Patient Combines Example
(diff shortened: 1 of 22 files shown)


daliboz:
Note: another outcome of this PR will be to update the template with the authorization types and the error link

content/millennium/dstu2/individuals/patient.md
```diff
@@ -71,11 +75,28 @@ Notes:
 - The `birthdate` parameter may be provided twice to indicate a date range, but must contain the inclusive prefixes 'le' and 'ge'
 - The `birthdate` parameter may be provided once with the following prefixes: 'ge', 'le', 'gt', 'lt', 'eq'
 
-### Response
+### Headers
+
+To successfully GET a Patient, the following headers must be provided. Authorization must be provided if accessing the closed endpoint.
+
+        Accept-Type: application/json+fhir
+        Authorization: <OAuth2 Bearer Token>
```
MaxPhilips: Should these be indented 4 spaces instead of 8 to match the formatted block in Request?

content/millennium/dstu2/individuals/patient.md
```diff
@@ -71,11 +75,28 @@ Notes:
...
+    GET https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Patient?_id=4342009
+
+#### Response
 
 <%= headers 200 %>
 <%= json(:dstu2_patient_bundle) %>
 
+### Errors
+
+The common [`errors`] may be returned.
```
MaxPhilips: Does the errors link need to be backticked?

content/millennium/dstu2/individuals/patient.md
```diff
@@ -86,12 +107,28 @@ _Implementation Notes_
...
+
+### Headers
+
+To successfully GET a Patient, the following headers must be provided. Authorization must be provided if accessing the closed endpoint.
+
+        Accept-Type: application/json+fhir
+        Authorization: <OAuth2 Bearer Token>
+
+### Example
+#### Request
```
MaxPhilips: Looks like this displays as hashes on the site, I think because there is no blank line between 121 and 122

content/millennium/dstu2/individuals/patient.md
daliboz: Probably :) Also need to update template for that.

content/millennium/dstu2/individuals/patient.md
daliboz: yeah...

content/millennium/dstu2/individuals/patient.md
daliboz: nope :)

content/millennium/dstu2/individuals/patient.md
```diff
@@ -71,11 +75,28 @@ Notes:
 - The `birthdate` parameter may be provided twice to indicate a date range, but must contain the inclusive prefixes 'le' and 'ge'
 - The `birthdate` parameter may be provided once with the following prefixes: 'ge', 'le', 'gt', 'lt', 'eq'
 
-### Response
+### Headers
```
dennispatterson: This section will essentially be the same across all resources and actions except for cases where OAuth token isn't used like in conformance-related resources.  I'm guessing we're just trying to be explicitly clear rather than relying on the Overview.  Rather than copy-pasting the whole block, we could have another function call to get the text that essentially just swaps out "GET a Patient" for other resources/actions.

daliboz: I made the existing headers method have a default case for this so that we're not copying the headers around. We have a lot of common text like this that we don't just generate, and it is something that starting from the template would help with. I'm not sure the text actually provides much... I could also make the default headers link back to the media type and auth sections on the overview, and remove the text intro all together...

dennispatterson: +1 to removing comment text and linking to other documentation for media type and auth headers

daliboz:
All changes made - let me know what you think. I'll expand the media type section on the overview for Binary once I get to that in the uplift...

lib/resources.rb
```diff
@@ -54,6 +69,13 @@ def link_header_rel(name, url)
         %Q{<#{url}>; rel="#{name}"}
       end
 
+      def default_headers()
```
dennispatterson: **super**-nit - no parens needed

dennispatterson:
+1 other than nit above

MaxPhilips:
+1

PR #123, opened 2017-03-31
Update documentation about contained resources in MedicationOrder and…

jfrhong:
… MedicationStatement.

### DIFF
--- a/content/millennium/dstu2/medications/medication-order.md
+++ b/content/millennium/dstu2/medications/medication-order.md
@@ -32,7 +32,7 @@ Search for MedicationOrders that meet supplied query parameters:
 
 _Notes_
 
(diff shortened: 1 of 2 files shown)


daliboz:
Clarify that we used contained when dealing with either multiple ingredient or compounded medications? The statement about them only existing within the context is confusing since we don't *always* use contained. IE: it's the concentrations/dilutions that typically cause it to be order specific.

cerbs:
+1

davischristi:
👍

jfrhong:
@daliboz how about "MedicationStatement.medication may be a reference to a contained Medication when the Medication cannot be represented by a CodeableConcept because it contains a unique combination of ingredients."

daliboz:
+1

PR #124, opened 2017-04-03
Update CarePlan text.div examples

MaxPhilips:
- Updated the implementation note about CarePlan description to say that the description is returned in text.div as ASCII text
- Updated text.div for the CarePlan entry and bundle examples
- Reformatted some of the example fields
- Removed 7 entries from the bundle example (kept the three most interesting CarePlans)


JFisk42:
+1

daliboz:
+1

PR #127, opened 2017-04-14
Add authorization header to Encounter actions

averyallen:
Updates Encounter page to include authorization header for each action.  Uses same format as [Patient](http://fhir.cerner.com/millennium/dstu2/individuals/patient/#authorization-types).  Encounters are only authorized for Practitioner and System.


daliboz:
+1

JFisk42:
+1

PR #154, opened 2017-07-27
Update Observation example to reflect flattened social history structure

MaxPhilips:
The intermediary 'form' social history is no longer displayed, and the value of the social history is its assessment.


daliboz:
+1 unless we can also add one of the results that's just assessment?

MaxPhilips:
I forgot that there was actually an assessment-only result already included.  I updated its id to include the 'a' it is now rendered with: 92f18b1

daliboz:
+1

MadhurRaj:
+1

PR #155, opened 2017-08-06
Add information about the embedded browser to SMART doc

kolkheang:
I'd like some feedback on the following additions.

@kpshek 
@daliboz 
@whitehatguy 
@mjhenkes 
@shriniketsarkar 
@zplata

### DIFF
--- a/content/smart.md
+++ b/content/smart.md
@@ -60,7 +60,9 @@ Apps should be designed to work in any browser for best user experience. For pro
 
 Cerner understands that it's counter-intuitive to prompt the user for their credential when launching a SMART application within *PowerChart*, where the user is already logged into the system. With that in mind, we designed the system to provide a good SSO experience for the users when any SMART application is launched within *PowerChart*. As a developer, you can help us give the users the best experience possible by following the following guideline when developing your apps to be embedded within *PowerChart*.
 
-- Due to technical limitations (see 'No Shared Cookies' section below), apps should use the same embedded browser (no pop-up) to authorize and launch when embedded in *PowerChart*. We still recommend using a pop-up for authorization and authentication in other workflows outside of *PowerChart*.
+- Due to technical limitations (see '[No Shared Cookies](#no-shared-cookies)' section below), apps should use the same embedded browser (no pop-up) to authorize and launch when embedded in *PowerChart*. We still recommend using a pop-up for authorization and authentication in other workflows outside of *PowerChart*.
+
+- If you'd like to flex your app based on the current execution context (embedded vs stand-alone mode), please see this [section](#embedded-in-powerchart) for more information.
 
 ### Embedded Browser Control ###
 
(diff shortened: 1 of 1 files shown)


content/smart.md
```diff
@@ -54,7 +54,29 @@ The recommended screen resolution varies based on the available screen's real es
...
+
+#### No Shared Cookies ####
+
+If your app uses a pop-up window for the authorization flow to allow users to authenticate with the system, that is fine for web and stand-alone apps.  However, if the app is going to be embedded within *PowerChart*, browser cookies will *NOT* be shared between the embedded browser and the newly pop-up browser window. What does this mean? It means that single-sign-on (SSO) will not work; the user would need to manually type in their credential when prompted, which is not the best experience for the users. It's best to perform all navigations and redirects using the embedded browser without needing to pop-up a new browser while embedding in *PowerChart*.  To learn more about the technical details of IE inner workings, please visit this [page](https://blogs.msdn.microsoft.com/ie/2008/03/11/ie8-and-loosely-coupled-ie-lcie/) and this [page](https://blogs.msdn.microsoft.com/ie/2010/03/04/tab-isolation/).
```
shriniketsarkar: Line 3 : embedded browser and the newly **_popped-up_** browser window.

content/smart.md
```diff
@@ -54,7 +54,29 @@ The recommended screen resolution varies based on the available screen's real es
...
+
+### Single Sign On (SSO) ###
+
+Cerner understands that it's counter-intuitive to prompt the user for their credential when launching a SMART application within *PowerChart*, where the user is already logged into the system. With that in mind, we designed the system to provide a good SSO experience for the users when any SMART application is launched within *PowerChart*. As a developer, you can help us give the users the best experience possible by following the following guideline when developing your apps to be embedded within *PowerChart*.
+
+- Due to technical limitations (see 'No Shared Cookies' section below), apps should use the same embedded browser (no pop-up) to authorize and launch when embedded in *PowerChart*. We still recommend using a pop-up for authorization and authentication in other workflows outside of *PowerChart*.
```
daliboz: Link to the section?

daliboz: Since we call out several times the difference between embedded vs. "normal" browser - is there a way we can document for an app to detect which environment it's running in (that wouldn't break other places they may launch from)?

daliboz: NM - see it below. We should link to that from other sections that would require that type of flexing.

daliboz:
otherwise +1

ericsornson:
+1

content/smart.md
```diff
@@ -54,7 +54,35 @@ The recommended screen resolution varies based on the available screen's real es
 
 ### Browser Requirement ###
 
-Apps should be designed to work in any browser for best user experience. For provider facing apps that run on the *Cerner Millennium* platform, Cerner currently supports only the embedded Internet Explorer browser (IE) at the moment. The minimum-supported embedded browser is IE 10. The latest supported browser varies based on the version of the browser that is currently installed on the device.
+Apps should be designed to work in any browser for best user experience. For provider facing apps that run on the *Cerner Millennium* platform, Cerner currently supports only the embedded Internet Explorer browser (IE) at the moment. The embedded browser control that we use is [IWebBrowser2 C++ interface](https://msdn.microsoft.com/en-us/library/aa752127(v=vs.85).aspx) by Microsoft. The minimum-supported embedded browser is IE 10. Apps *must* be designed to work with IE 10. The latest supported browser varies based on the version of the browser that is currently installed on the device.
```
mjhenkes: You could mention that Microsoft offers a free IE 10 VM for application testing prior to embedding. https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

kolkheang: Ah, yes.  I forgot I mentioned this in the other guide.

content/smart.md
```diff
@@ -54,7 +54,37 @@ The recommended screen resolution varies based on the available screen's real es
...
```
kpshek: To me, this sounds like we only render in IE 10 rather than IE 10 is the minimum, but other more recent versions of IE (eg, 11 & Edge) are also used. Can we make that more explicit here?

content/smart.md
```diff
@@ -54,7 +54,37 @@ The recommended screen resolution varies based on the available screen's real es
...
+
+If you'd like to flex your app based on the current execution context (embedded vs stand-alone mode), please see this [section](#embedded-in-powerchart) for more information.
+
+#### HTML5 Session Storage ####
+
+With the introduction of HTML5, [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) is a good way to persist data within a browser's session. If your app is going to be embedded within *PowerChart*, we do *NOT* recommend using `sessionStorage`. The reason is because due to our technical implementation of the embedded browser, the session is not sandboxed to per "tab" like the documentation states. With our implementation of the embedded browser, the session is shared across all instances of the embedded browser. This in turn, would leak the session to other "tabs".
```
kpshek: fyi - With 2018.01, our embedded IE control will no longer have this sessionStorage behavior since we're spawning each browser control in its own thread. I think this advice is still sound but perhaps we should state that depending on the Millennium release, sessionStorage may not be isolated per instances of the embedded browser (and thus web pages) and as such, we do not recommend using it.

kolkheang:
@kpshek  - updated

PR #157, opened 2017-08-09
Apply template to encounter, fix links in Immunization

daliboz:
* Applied the official template to the Millennium Encounter resource
* Fixed some Immunization links I noticed went to latest instead of dstu2


MaxPhilips:
+1

PR #158, opened 2017-08-09
Added auth and applied template to Millennium Person

daliboz:
Changes to Millennium Person:
* Added auth types accepted
* Applied template


MaxPhilips:
+1

PR #159, opened 2017-08-09
Add auth types and template to Millennium RelatedPerson

daliboz:
Added supported authentication and applied template to Millennium RelatedPerson


MaxPhilips:
+1

PR #166, opened 2017-08-25
Update AllergyIntolerance and Condition Note

averyallen:
Adds notes to AllergyIntolerance.note and Condition.notes.


ericsornson:
+1

daliboz:
Instead of adding it in the body fields (since it already has note.x fields, and they may completely skip that), how about adding it to the Notes section just above the body fields table?

averyallen:
@daliboz, moved to the Notes section above the field tables:  0a2f360

daliboz:
+1

PR #167, opened 2017-08-29
Add default and maximum _count to MedicationOrder

MaxPhilips:


MaxPhilips:
Closing due to offline discussion

cravelight:
Did the offline conversation lead the team to not publish this?  If so, what was the logic?

This commit would have saved me 20 minutes of research trying to figure out why MedOrders were maxing out at 60 when we have _count set to 100.

As a side note - the first person to push the max up to 200 gets a free steak dinner from Juxly.  ;)

MaxPhilips:
We've previously run into a situation where underlying services (i.e. MedOrder retrieval) have changed their default/max counts, outdating doc we published to the site.  Additionally, we plan to continuously tune the precise numbers based on service performance, so keeping them up to date may involve quite a few changes.

Personally, I think it would be worth the pain of maintenance to have them documented in a public facing spot, but we are going forward with the strategy of not publishing them at this time.

In general, it's safe to assume that a consistent lesser amount of results returned than the count you requested is behind the scenes page throttling.

PR #168, opened 2017-09-01
Added Code Portal Button

beichhor:
To provide more visibility to the Code Portal, a prominent button is displayed on the Millennium, Authorization, and SMART sections of the fhir.cerner.com site.

### DIFF
--- a/Rules
+++ b/Rules
@@ -19,6 +19,7 @@
 preprocess do
   add_version_attribute
   add_solution_attribute
+  add_button_attribute
   RedirectGenerator.generate(config[:redirects], @items)
 end
 
--- a/static/css/documentation.css
+++ b/static/css/documentation.css
@@ -218,6 +218,12 @@ a img {
   position: relative;
 }
 
+#header-button {
+  background: #5C9A32;
+  margin: 10px 0 0;
+  float: right;
+}
+
 #top-header {
   height: 50px;
   padding: 15px 0;
--- a/layouts/header.html
+++ b/layouts/header.html
@@ -4,6 +4,9 @@
     <div class="header-wrapper">
 
       <a class="logo" href="https://code.cerner.com">Cerner Corporation</a>
+      <% if @item[:header_button] %>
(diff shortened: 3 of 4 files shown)


static/css/documentation.css
```diff
@@ -218,6 +218,12 @@ a img {
   position: relative;
 }
 
+#header-button {
+  background: #5C9A32;
```
MaxPhilips: May want to use #7BC143

MaxPhilips:
+1

static/css/documentation.css
MaxPhilips: Hm, I'm wavering back to the original color.  I think it looks better a bit darker

beichhor: I think the original color definitely makes the text easier to read.

beichhor: I've gone ahead and dropped the change to revert it back to it's previous color.

Rules
```diff
@@ -19,6 +19,7 @@
 preprocess do
   add_version_attribute
   add_solution_attribute
+  add_button_attribute
```
daliboz: May be simpler to just check @item[:solution] != 'soarian' in the header layout? That way, if we add other solutions, it's obvious when this piece depends on solution?

static/css/documentation.css
daliboz: looks good - goal was to keep it consistent with the code site, and this color matches.

daliboz:
+1 other than possibly simplifying to use existing attributes :)

Rules
beichhor: The only issue with that is that the button will show up on the homepage. Not sure if that is expected behavior you would want.

daliboz: eh. probably not. Darn.

daliboz:
Ok, just +1 :)

shriniketsarkar:
This update was long overdue. I am glad we are getting this done :) Thanks @beichhor .

PR #177, opened 2017-11-21
Update authorization.md

PeterLamar:
The documentation misses a key detail that is shared in this google group comment by a Cerner representative. 

https://groups.google.com/forum/#!searchin/cerner-fhir-developers/metadata|sort:date/cerner-fhir-developers/86XQJ5xktFA/J6ORygVeCAAJ

Namely that the /metadata needs the extension ?_format=json to work. Adding this to the documentation will hopefully save future devs some effort.

### DIFF
--- a/content/authorization.md
+++ b/content/authorization.md
@@ -115,7 +115,10 @@ FHIR<sup>®</sup> conformance document.
 
 The FHIR<sup>®</sup> Conformance resource can be retrieved by performing
 an HTTP GET against the resource located at the relative path
-of ``./metadata`` from the FHIR<sup>®</sup> base URL.
+of ``./metadata?_format=json`` from the FHIR<sup>®</sup> base URL. An
+example of this API call is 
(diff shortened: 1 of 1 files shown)


MaxPhilips:
Thanks for opening this pull request @PeterLamar 

To make a successful metadata call (like other calls to Cerner's FHIR server), either the _format query parameter needs to be set or an "Accept: application/json+fhir" header needs to be set.

An example metadata call (along with more information about accept types) using the header strategy is found on this page: http://fhir.cerner.com/millennium/dstu2/

The Authorization page contains documentation on authorization, not general server calls so I would lean towards not adding an example here.  I've logged an internal task to evaluate this point of confusion so we can make sure everything is easy to consume.  I'll update this thread with the results.

daliboz:
@PeterLamar @MaxPhilips I added some changes that should help make the whole metadata concept better documented for everyone. Fixed an issue I found while doing this as well. Let me know what you think.

MaxPhilips:
+1, the changes look good to me

PR #180, opened 2017-11-30
Documentation for Person combines

cerbs:
Updating the Person resource to address upcoming person combine behavior similar to the Patient resource.

### DIFF
--- a/content/millennium/dstu2/entities/person.md
+++ b/content/millennium/dstu2/entities/person.md
@@ -95,5 +95,27 @@ List an individual Person by its id:
 
 The common [errors] may be returned.
 
+### Person Combines Example
+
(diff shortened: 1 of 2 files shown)


ericsornson:
Millennium uses the term `combine`, but FHIR refers to it as `merging`.

Should we be using the term `merging` in this documentation except where we refer to the millennium operation?  Might be worth calling out that distinction here as well.

cerbs:
I would prefer to stay with "combine" to be consistent with the existing Patient resource documentation. This is also describing how a Millennium "combine" is reflected through our API.

beichhor:
+1

averyallen:
+1

PR #181, opened 2017-12-14
Fix links to application and code.cerner.com

daliboz:
Tried to find all links to the old application and code.cerner.com to correct them after the site was reorganized.


ericsornson:
+1

PR #189, opened 2018-01-11
Add Documentation for StructureDefinition

averyallen:
Adds documentation for StructureDefinition.  

A couple things I could use feedback on:

- Do we want to list the custom extensions for which we define StructureDefinitions?  We do something similar in OperationDefinition.
- If so, I could use some help thinking of appropriate descriptions for each of the extensions.  I've gone ahead and included them and taken a shot at describing them.

### DIFF
--- a/content/millennium/dstu2/structure-definition.md
+++ b/content/millennium/dstu2/structure-definition.md
@@ -1,0 +1,75 @@
+---
+title: StructureDefinition | DSTU 2 API
+---
+
+# StructureDefinition
+
+* TOC
+{:toc}
+
+## Overview
+
+The StructureDefinition resource describes a FHIR structure including data elements and their usage. Our current implementation uses this resource to define custom extensions.
+
+The following fields are returned if valued:
+
+* [Id](http://hl7.org/fhir/dstu2/resource-definitions.html#Resource.id){:target="_blank"}
+* [Publisher](http://hl7.org/fhir/DSTU2/structuredefinition-definitions.html#StructureDefinition.publisher){:target="_blank"}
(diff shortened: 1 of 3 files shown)


beichhor:
+1

daliboz:
I like the way the list is built in the current version - the link to the "source of truth" for documentation is nice. It let's the extension be described in context of the impacted resource. If we ever had a very common extension, we could define it here and then link from the other resources.

+1

content/millennium/dstu2/structure-definition.md
```diff
@@ -0,0 +1,75 @@
...
+* [Kind](http://hl7.org/fhir/DSTU2/structuredefinition-definitions.html#StructureDefinition.kind){:target="_blank"}
+* [Differential](http://hl7.org/fhir/DSTU2/structuredefinition-definitions.html#StructureDefinition.differential){:target="_blank"}
+* [Abstract](http://hl7.org/fhir/DSTU2/structuredefinition-definitions.html#StructureDefinition.abstract){:target="_blank"}
+* [Base](http://hl7.org/fhir/DSTU2/structuredefinition-definitions.html#StructureDefinition.base){:target="_blank"}
+* [Name](http://hl7.org/fhir/DSTU2/structuredefinition-definitions.html#StructureDefinition.name){:target="_blank"}
+
+_List of StructureDefinitions_
+
+* [`patient-friendly-display`](http://fhir.cerner.com/millennium/dstu2/medications/medication-order/#custom-extensions): Display string suitable for patient viewing.
+* [`scheduling-location`](http://fhir.cerner.com/millennium/dstu2/scheduling/schedule/#custom-extensions): Reference to the location an appointment being scheduled.
```
averyallen: Just noticed a typo here. 'Reference to the location an appointment being scheduled.' 
Should be updated to: 'Reference to the location **of** an appointment being scheduled.'

PR #191, opened 2018-01-19
Add Documentation for MedicationOrder _lastUpdated Parameter

averyallen:
Adds _lastUpdated to params section.  Also adds a notes section to describe _lastUpdated parameter requirements.

### DIFF
--- a/content/millennium/dstu2/medications/medication-order.md
+++ b/content/millennium/dstu2/medications/medication-order.md
@@ -86,6 +86,7 @@ _Implementation Notes_
 
  - The `-timing-boundsPeriod` and `_lastUpdated` parameters may not be provided at the same time.
  - The `_lastUpdated` parameter must have a time, may be provided up to two times, and must use the `ge` or `le` prefixes.  When provided twice, the lower value must have the `ge` prefix and the higher value must have the `le` prefix.
+ - Searching with the `_lastUpdated` parameter will only detect changes to fields that affect the clinical meaning of the order. An example of the types of changes that won't be caught by this query are changes that would affect the version, but not the FHIR content.
 
 ### Headers
 
--- a/content/millennium/dstu2/medications/medication-order.md
+++ b/content/millennium/dstu2/medications/medication-order.md
@@ -9,7 +9,7 @@ title: MedicationOrder | DSTU 2 API
 
 ## Overview
 
-The Medication Order resource provides orders for all medications along with administration instructions for a patient in both the inpatient and outpatient setting (orders/prescriptions filled by a pharmacy and discharge medication orders). This resource does NOT include historical or home medications documented or reported to have been taken by patient, significant other or another provider. 
+The Medication Order resource provides orders for all medications along with administration instructions for a patient in both the inpatient and outpatient setting (orders/prescriptions filled by a pharmacy and discharge medication orders). This resource does NOT include historical or home medications documented or reported to have been taken by patient, significant other or another provider.
 
 If the Medication Order represents a prescription (something the patient takes at home), the start, stop, and other data may not be a representation of when the medication was taken. For example, the system may not know if the patient ever filled or took the prescribed medication, or when the prescription was filled.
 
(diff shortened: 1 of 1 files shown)


beichhor:
+1

content/millennium/dstu2/medications/medication-order.md
```diff
@@ -79,8 +79,17 @@ _Implementation Notes_
  `patient`                 | Y         | [`reference`] | The identity of a patient to list orders for. Example: `12345`
  `status`                  | N         | [`token`]     | The status of the medication order, may be a list separated by commas. Example: `active,draft`
  `-timing-boundsPeriod`    | N         | [`date`]      | The date-time which should fall within the [period] the medication should be given to the patient. Must be prefixed by 'ge'. Example: `ge2014-05-19T20:54:02.000Z`
+ `_lastUpdated`            | N         | [`date`]      | The date-time within which the most recent clinically relevant update was made to the medication. Must be prefixed by 'ge' or 'le'. Example: `ge2014-05-19T20:54:02.000Z`
```
MaxPhilips: I think 'within' is confusing as written for single instances of _lastUpdated.  What about something like:

"An explicit or implied date-time range within which the most recent clinically relevant update was made to the medication."

content/millennium/dstu2/medications/medication-order.md
```diff
@@ -79,8 +79,17 @@ _Implementation Notes_
...
  [`_count`]                | N         | [`number`]    | The maximum number of results to include in a page. Example: `50`
 
+ Notes:
+
+ - The `-timing-boundsPeriod` and `_lastUpdated` parameters may not be provided at the same time.
+ - When the `_lastUpdated` parameter is provided:
+   - Up to two dates-times may be provided.
```
MaxPhilips: date-times

I would be clear here about how they can be provided (comma separated vs. two instances of the parameter)

averyallen: How about this:

```
- When using the `_lastUpdated` parameter:
   - It may be provided twice with each occurrence set to a single date-time.
   - Each date must have a time component and be prefixed by `ge` or `le`.
   - When provided twice, the earliest date-time must have the `ge` prefix and the latest date-time must have the `le` prefix.
```

MaxPhilips: In a couple other resources, we do a single bullet instead of breaking it out

Observation/CarePlan:
The XXXXX parameter may be provided up to two times, and must use the ge or le prefixes. When provided twice, the lower value must have the ge prefix and the higher value must have the le prefix.

Procedure:
The XXXXX parameter must have a time, may be provided up to two times, and must use the ge or le prefixes. When provided twice, the lower value must have the ge prefix and the higher value must have the le prefix.

I think copying the language for we have for Procedure would be good

daliboz:
I believe the date MUST have a time/zone component, correct? If so, we need to document that requirement.

daliboz:
Also, I believe this only looks for clinically significant updates, which we may want to mention.

averyallen:
Yep, the time/zone component is required.  Hoping that should be covered with the description here: https://github.com/cerner/fhir.cerner.com/pull/191/files#diff-bc268db264ad46dc39fdd6e2c2842436R88

Also we explicitly state clinically relevant updates here: https://github.com/cerner/fhir.cerner.com/pull/191/files#diff-bc268db264ad46dc39fdd6e2c2842436R82

Do you think we need to add a note as well to make sure it doesn't go missed?

@daliboz

daliboz:
Yes, I think we'll need a not to explain what "clinically relevant" means in a notes section :) We should also double-check with the local team to understand what fields would trigger a change in the FHIR resource (other than last updated) but not a change in the content. I'm guessing comments is one, but not sure.

MaxPhilips:
+1 for comments I added above

+1 for PR after resolution of clinically relevant note

averyallen:
@daliboz, added the clinically relevant note.  Think we are good to merge if everything looks okay to you.

daliboz:
+1

PR #194, opened 2018-02-27
Patient Write Documentation

MadhurRaj:


daliboz:
Apparently, we also need to add that we support the care providers field...

https://groups.google.com/d/msg/cerner-fhir-developers/tLNiRarnOdY/_UeShfFFAQAJ

Maybe we need to do a delta of the current code w/the current doc? Then update the limitations on the create doc to ensure it covers all the gaps between what read supports vs write.

lib/resources/dstu2/patient.yaml
```diff
@@ -27,15 +29,282 @@ fields:
...
+  required: 'Yes'
+  cardinality: 1..*
+  type: Identifier
+  description: The identifiers for the patient.
+  action: create
+  example: |
+    {
+      "identifier": [
+        {
+          "assigner": {
```
daliboz: I thought we also required the identifier type to come in?

MadhurRaj: We do on everything but the first. The first can only contain the assigner.

MadhurRaj:
Added Care Provider and Multiple Birth Information fields to the overview

lib/resources/dstu2/patient.yaml
```diff
@@ -27,15 +29,282 @@ fields:
...
+  required: 'No'
+  cardinality: 0..1
+  type: date
+  description: The date of birth for the individual.
+  action: create
+  example: |
+    {
+      "birthDate": "1990-09-15"
+    }
+  note: birthDate must not contain a time component.
```
daliboz: I don't think we need this note: birthDate is defined as a "Date" field, it's impossible for it to contain a time (that would be DateTime or instant types) :)

content/millennium/dstu2/individuals/patient.md
```diff
@@ -145,6 +147,88 @@ The ability to perform patient combine or uncombine operations is not available
...
 
+## Create
+
+Create an individual Patient:
+
+    POST /Patient
+
+_Implementation Notes_
+
+* The following elements are not supported and will not be used while creating the patient:
```
daliboz: Maybe clarify to "and will cause an error to be returned if set" instead of stating they'll not be used. It makes it sound like we'll ignore them :)

content/millennium/dstu2/individuals/patient.md
```diff
@@ -145,6 +147,88 @@ The ability to perform patient combine or uncombine operations is not available
...
+      Content-Length → 0
+      Content-Type → application/json
+  </pre>
+
+  The `ETag` response header indicates the current `If-Match` version to use on subsequent updates.
+
+
+### Errors
+
+The common [errors] may be returned. In addition, [OperationOutcomes] may be returned in the following scenarios:
```
daliboz: Need to fill in the OperationOutcomes link, it just shows up with no link and brackets right now

daliboz:
+1 other than comments

content/millennium/dstu2/individuals/patient.md
```diff
@@ -145,6 +149,88 @@ The ability to perform patient combine or uncombine operations is not available
...
+
+### Authorization Types
+
+<%= authorization_types(practitioner: true, system: true) %>
+
+### Headers
+
+<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', Accept: 'application/json+fhir', 'Content-Type': 'application/json+fhir'} %>
+
+### Body fields
```
MaxPhilips: I know we have lowercase 'fields' elsewhere on the site, but this is a title and should be uppercase. Not sure if consistency or correctness is more aesthetically pleasing

content/millennium/dstu2/individuals/patient.md
```diff
@@ -145,6 +149,88 @@ The ability to perform patient combine or uncombine operations is not available
...
+
+Notes:
+
+* Birth Sex may be recorded as an extension.
```
MaxPhilips: Could this be a note in the table? Or is it here to draw more attention?

content/millennium/dstu2/individuals/patient.md
MadhurRaj: Hmm....I'd prefer to go for correctness I think. I'll update this to Fields

content/millennium/dstu2/individuals/patient.md
MadhurRaj: I did intend to call it out here because it isn't listed as a field on the FHIR spec. I can remove it if it looks awkward here.

MaxPhilips:
+1 pending outstanding comments

content/millennium/dstu2/individuals/patient.md
MadhurRaj: Added the link.

lib/resources/dstu2/patient.yaml
MadhurRaj: Removed the note

content/millennium/dstu2/individuals/patient.md
MadhurRaj: updated the note to 'will cause an error to be returned if set'

PR #202, opened 2018-07-17
Active status not supported by DSTU2 Specification

smoyssidis:
"status=active" not supported by DSTU2 Specification:
http://hl7.org/fhir/DSTU2/valueset-medication-admin-status.html


daliboz:
+1, Thanks @smoyssidis !

PR #204, opened 2018-07-26
Add DSTU2 documentation for Appointment status update

mp04:
Updates Appointment resource to include update operation.

### DIFF
--- a/content/millennium/dstu2/scheduling/appointment.md
+++ b/content/millennium/dstu2/scheduling/appointment.md
@@ -268,6 +268,11 @@ The `ETag` response header indicates the current `If-Match` version to use on su
 
 The common [errors] and [OperationOutcomes] may be returned.
 
+In addition, the following error may be returned:
+
+* Updating an Appointment resource without sending the `If-Match` header will result in a `412 Precondition Failed` response.
+
+
 [`reference`]: http://hl7.org/fhir/DSTU2/search.html#reference
 [`token`]: http://hl7.org/fhir/DSTU2/search.html#token
 [`date`]: http://hl7.org/fhir/DSTU2/search.html#date
--- a/content/millennium/dstu2/scheduling/appointment.md
+++ b/content/millennium/dstu2/scheduling/appointment.md
@@ -197,6 +197,71 @@ _Implementation Notes_
 
 The `ETag` response header indicates the current `If-Match` version to use on subsequent updates.
 
+## Update
+
+Update an Appointment.
+
+    PUT /Appointment/:id
+
+_Implementation Notes_
+
+* The only supported change is to update the [Appointment.status] to either `arrived` or `cancelled`.
+* `Appointment.participant` must have at least one participant.
+* `Appointment.participant.status` must be `accepted` for each participant
+
+### Authorization Types
+
+<%= authorization_types(practitioner: true, patient: false, system: true) %>
+
+### Headers
+
+<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', Accept: 'application/json+fhir',
+                   'Content-Type': 'application/json+fhir', 'If-Match': 'W/"&lt;Current version of the Appointment resource>"'} %>
+
+### Body fields
+
+<%= definition_table(:appointment, :update, :dstu2) %>
+
+### Example
+
+#### Request
+
+    PUT https://fhir-ehr.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Appointment/1620549
+
+#### Body
+
+<%= json(:dstu2_appointment_update) %>
+
+#### Response
+
+<%= headers status: 200 %>
+<pre class="terminal">
+    Connection → Keep-Alive
+    Content-Encoding → gzip
+    Content-Length → 20
+    Content-Type → text/html; charset=UTF-8
+    Date → Wed, 13 Jan 2016 21:50:53 GMT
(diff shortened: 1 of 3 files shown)


lib/resources/example_json/dstu2_examples_appointment.rb
```diff
@@ -412,5 +412,26 @@ module Resources
       "status": "proposed"
     }
 
+    DSTU2_APPOINTMENT_UPDATE ||= {
```
JFisk42: For this example lets use a result from the sample search provided for Appointment search (https://fhir-open.sandboxcerner.com/dstu2/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Appointment?date=2017&patient=4704007)

JFisk42: Also it looks like a lot of the formatting is off here. Not sure how it would look on the site but I imagine it wouldn't look right.

mp04: Addressed. Used actual result and fixed formatting.

lib/resources/example_json/dstu2_examples_appointment.rb
```diff
@@ -404,13 +404,28 @@ module Resources
...
+            "reference": "Patient/4704007",
+            "display": "Smart, Barney R"
+          },
+          "required": "required",
+          "status": "accepted"
+        }
+      ],
+      "status": "arrived"
+    }
+
```
JFisk42: Nitpick: extra line

JFisk42:
+1

content/millennium/dstu2/scheduling/appointment.md
```diff
@@ -197,6 +197,71 @@ _Implementation Notes_
...
+
+Update an Appointment.
+
+    PUT /Appointment/:id
+
+_Implementation Notes_
+
+* The only supported change is to update the [Appointment.status] to either `arrived` or `cancelled`.
+* `Appointment.participant` must have at least one participant.
+* `Appointment.participant.status` must be `accepted` for each participant
```
MaxPhilips: Add a period so all the notes are full sentences

lib/resources/example_json/dstu2_examples_appointment.rb
```diff
@@ -404,13 +404,28 @@ module Resources
...
-            "display": "PATIENT, TEST"
+            "display": "Smart, Barney R"
           },
           "status": "needs-action"
         }
       ],
       "status": "proposed"
     }
 
+    DSTU2_APPOINTMENT_CREATE ||= {
```
MaxPhilips: Rename to UPDATE otherwise site does not generate

content/millennium/dstu2/scheduling/appointment.md
daliboz: Probably an "all other fields will be ignored" or similar statement? We don't check for changes, correct?

daliboz:
In the update, it says only the appointment status can be changed, and mentions that the participants have to be in a specific state. I assume that state can't/shouldn't change from what it is on the initial appointment?

mp04:
@daliboz, that's right.
The participants' status won't change. Only the appointment status will be updated.
I believe that the workflow requires the participants need to be in a specific (accepted) state for the appointment to be cancelled or checked-in.

content/millennium/dstu2/scheduling/appointment.md
```diff
@@ -197,6 +197,72 @@ _Implementation Notes_
...
+    x-request-id → 9dba8326-899a-406f-a125-3fc3d6605dad
+    x-runtime → 0.653722
+    x-xss-protection → 1; mode=block
+</pre>
+
+The `ETag` response header indicates the current `If-Match` version to use on subsequent updates.
+
 ### Errors
 
 The common [errors] and [OperationOutcomes] may be returned.
```
MaxPhilips: Add a note that skipping If-Match will result in 412

lib/resources/dstu2/appointment.yaml
```diff
@@ -69,6 +69,19 @@ fields:
...
+  required: 'Yes'
+  cardinality: 1..1
+  type: code
+  description: A code representing the overall status of the Appointment. Each of the participants has their own participation status which indicates their involvement in the process, however this status indicates the shared status.
+  note: When updating an appointment, the status must be "arrived" or "cancelled".
+  action: update
+  example: |
+      {
+        "status": "arrived"
+      }
```
MaxPhilips: Nit: indent here and L138-148 only 2 spaces

content/millennium/dstu2/scheduling/appointment.md
mp04: Addressed in 3a3f2aa

MaxPhilips:
+1

JFisk42:
In addition to the normal errors, we also introduced the possibility of igneous_server erroring with "423 Locked" and "424 Failed Dependency". Do we need to add these into the documentation? @daliboz @MaxPhilips

daliboz:
Yes, those should be added to this resource's specific errors (and the OperationOutcome that's returned with them should be documented as well on this resource page)

mp04:
Comment addressed. I've add 423 and 424 to the additional errors.
Not sure if what I have for 424 is the best way to represent it.
I currently have,

> If the Appointment resource could not be updated because of an operation that is necessary for the update (eg. encounter association), `424 Failed Dependency` response will be returned.

PR #207, opened 2018-08-15
Update DSTU2 documentation for Slot search.

mp04:
Updates DSTU2 documentation for Slot search.
slot-type is no longer a required parameter and multiple slot-types can be provided when searching for slots.

**Note:** Do not merge until the changes are released.

### DIFF
--- a/content/millennium/dstu2/scheduling/slot.md
+++ b/content/millennium/dstu2/scheduling/slot.md
@@ -73,12 +73,9 @@ _Implementation Notes_
 
 Notes:   
 
+- At least one of the `slot-type`, `schedule.actor`, and `-location` parameters must be provided.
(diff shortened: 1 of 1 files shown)


content/millennium/dstu2/scheduling/slot.md
```diff
@@ -73,12 +73,9 @@ _Implementation Notes_
 
 Notes:   
 
+- At least one of the `slot-type`, `schedule.actor`, and `-location` parameters must be provided.
 - The `slot-type`, `schedule.actor`, and `-location` parameters may be included only once.
```
JFisk42: There needs to be language that clarifies that slot-type can only be included once, but a comma separated series of slot-types is allowed now.

PR #222, opened 2018-12-21
Slot proprietary codes

ericsornson:
adding documentation for slot search proprietary codes
- new slot type search param using  proprietary codes
- update output response which includes proprietary codes

### DIFF
--- a/content/millennium/dstu2/scheduling/slot.md
+++ b/content/millennium/dstu2/scheduling/slot.md
@@ -81,6 +81,10 @@ Notes:
 * When `slot-type` is provided, `start` must be provided.
   For example, `start=2018&slot-type=http://snomed.info/sct|394581000,http://snomed.info/sct|394602003`
 
+* The `slot-type` parameter may contain standard codes or proprietary codes.
+  * When using standard codes, choose codes bound to Slot.type as noted in the [Terminology Bindings](#terminology-bindings) table above.
+  * When using proprietary codes, the system should be `https://fhir.cerner.com/<your EHR source id>/codeSet/<code set>` (where code set is a Millennium code set) and the code should be a Millennium code value. Example: `https://fhir.cerner.com/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/codeSet/14249|4062744`
+
 * The `start` parameter may be provided:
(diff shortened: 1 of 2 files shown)


content/millennium/dstu2/scheduling/slot.md
```diff
@@ -63,32 +63,36 @@ _Implementation Notes_
...
-- When `slot-type` is provided, `start` must be provided.
+* When `slot-type` is provided, `start` must be provided.
   For example, `start=2018&slot-type=http://snomed.info/sct|394581000,http://snomed.info/sct|394602003`
 
-- The `start` parameter may be provided:
-  - once without a prefix or time component to imply a date range. (e.g. `&start=2016`, `&start=2016-07`, `&start=2016-07-04`)
-  - twice with the prefixes `ge` and `lt` to indicate a specific range. The date and prefix pairs must define
+* The `slot-type` parameter may contain standard codes or proprietary codes.
+  * When using standard codes, choose codes bound to Slot.type as noted in the [Terminology Bindings](#terminology-bindings) table above.
+  * When using proprietary codes, the system should be `https://fhir.cerner.com/<your EHR source id>/codeSet/<code set>` (where code set is a Millennium code set) and the code should be a Millennium code value. Example: `https://fhir.cerner.com/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/codeSet/14249|4062744`
```
praneethk2094: Can we switch this to reflect 'start' needs to exist when slot-type is provided, something like 
`start=2018&slot-type=https://fhir.cerner.com/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/codeSet/14249|4062744`

Also, i think we should document that we only support one slot-type at once, either proprietary or standard.

MaxPhilips: disagree with the first part - this example is only the value of a slot-type parameter. this is not a full blown query. the note about slot-type being provided mandating start must be provided already exists on L81

good catch for the second. update L84 to "* The `slot-type` parameter may contain standard codes or proprietary codes. When using standard codes, proprietary codes may not be provided, and vice versa."

ericsornson: good call, fixed - 647ac49d650c225fb918808f2d2df0f5d686d51e

ericsornson: ugh, let me fix typos

PR #242, opened 2019-03-28
[R4] Add documentation for Patient Patch capability

zsura:
Adds R4 documentation for Patient Patch capability.

### DIFF
--- a/content/millennium/r4/individuals/patient.md
+++ b/content/millennium/r4/individuals/patient.md
@@ -210,13 +210,69 @@ The `ETag` response header indicates the current `If-Match` version to use on su
 
 The common [errors] and [OperationOutcomes] may be returned.
 
+## Patch
+
+Patch an existing Patient.
+
+    PATCH /Patient/:id
+
+_Implementation Notes_
+
+* Only operations on the paths listed below are supported.
+* When a path includes `{index}`, it refers to the index of the element in the list of elements. The indexes here are 0 based.
+
+### Authorization Types
+
+<%= authorization_types(practitioner: true, system: true) %>
+
+### Headers
+
+<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', Accept: 'application/fhir+json',
+                   'Content-Type': 'application/json-patch+json', 'If-Match': 'W/"&lt;Current version of the Patient resource>"'} %>
+
+### Patch Operations
+
+<%= patch_definition_table(:patient_patch, :r4) %>
+
+### Example
+
+#### Request
+
+    PATCH https://fhir-ehr.sandboxcerner.com/r4/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Patient/5786010
+
+#### Body
+
+<%= json(:r4_patient_patch) %>
+
+### Response
+
+<%= headers status: 200 %>
+<pre class="terminal">
+    Cache-Control: no-cache
+    Content-Length: 0
+    Content-Type: application/fhir+json
+    Date: Wed, 27 Mar 2019 17:23:14 GMT
+    Etag: W/"1"
+    Last-Modified: Wed, 27 Mar 2019 17:23:14 GMT
+    Server-Response-Time: 296.405243
+    Status: 200 OK
+    Vary: Origin
+    X-Request-Id: 81c05ba01b2c19e5bf421449ff8e97eb
+    X-Runtime: 0.296306
+</pre>
+
+The `ETag` response header indicates the current `If-Match` version to use on subsequent updates.
+
+### Errors
+
+The common [errors] and [OperationOutcomes] may be returned.
+
 [`_count`]: http://hl7.org/fhir/R4/search.html#count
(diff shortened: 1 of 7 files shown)


lib/resources/r4/patient_patch.yaml
```diff
@@ -0,0 +1,760 @@
...
+    [
+      {
+        "path": "/identifier/1/id",
+        "op": "test",
+        "value": "CI-9926250-1"
+      }
+    ]
+- name: remove-identifier
+  path: /identifier/{index}
+  operation: remove
```
MaxPhilips: cardinality and type?

couple others don't have these keys either

lib/resources/r4/patient_patch.yaml
```diff
@@ -0,0 +1,760 @@
+---
+name: Patient
+field_name_base_url: https://hl7.org/fhir/R4/patient-definitions.html#Patient
+operations:
+- name: add-identifier
+  path: /identifier/-
+  operation: add
+  cardinality: 1..1
```
MaxPhilips: I think the cardinality for all these should actually be 0..*

all the operations can be provided many times, and all are mostly optional right? some tests are required when other ops are present

content/millennium/r4/individuals/patient.md
```diff
@@ -208,14 +208,69 @@ The `ETag` response header indicates the current `If-Match` version to use on su
...
+
+#### Body
+
+<%= json(:r4_patient_patch) %>
+
+### Response
+
+<%= headers status: 200 %>
+<pre class="terminal">
+    Cache-Control: no-cache
```
MaxPhilips: remove leading spaces inside pre tag

lib/resources/r4/patient_patch.yaml
zsura: So cardinality here is for the value in the patch operation and not the actual operation.
Might make it more clearer if the headers are updated to `Value Cardinality` and `Value Type`.

lib/resources/r4/patient_patch.yaml
zsura: So for remove operations a value is not required. Made sense to not set the cardinality/type instead of setting it to `N/A`

lib/resources/r4/patient_patch.yaml
zsura: Updated the headers with https://github.com/cerner/fhir.cerner.com/pull/242/commits/ccc35fc82a441599bcc1e8f2333a0a2324293ef8

content/millennium/r4/individuals/patient.md
zsura: Done.
https://github.com/cerner/fhir.cerner.com/pull/242/commits/ccc35fc82a441599bcc1e8f2333a0a2324293ef8

zsura:
As part of https://github.com/cerner/fhir.cerner.com/pull/242/commits/ccc35fc82a441599bcc1e8f2333a0a2324293ef8 i have updated the examples to remove the patch operation from the enclosing array.

lib/resources/r4/patient_patch.yaml
MaxPhilips: having blank boxes looks to me like we forgot something, I would prefer an N/A

lib/resources/r4/patient_patch.yaml
```diff
@@ -0,0 +1,682 @@
...
+  description: Replace the `Identifier.system` of the identifier at the given <code>{index}</code> in the list of patient identifiers.
+  example: |
+    {
+      "path": "/identifier/1/system",
+      "op": "replace",
+      "value": "urn:oid:2.16.840.1.113883.3.13.6"
+    }
+  note: >
+    <ul>
+      <li>Requires a <a href="#test-identifier-id">test operation</a> to be provided for the `Identifier.id` of the identifier that's `system` is intended to be replaced.</li>
```
ebaranowski: The "that's" need to be changed out to "whose"

lib/resources/r4/patient_patch.yaml
```diff
@@ -0,0 +1,682 @@
...
+      "value": {
+        "start": "2016-01-02T00:00:00-05:00",
+        "end": "2022-01-02T00:00:00-05:00"
+      }
+    }
+  note: >
+    <ul>
+      <li>Requires a <a href="#test-identifier-id">test operation</a> to be provided for the `Identifier.id` of the identifier that's `period` is intended to be replaced.</li>
+      <li>The `start` and `end` when provided must include a time component with a timezone.</li>
+      <li>When `start` or `end` is not provided its value will be unset.</li>
```
ebaranowski: Nit:

should be 'are' instead of 'is'

zsura:
Rebased over latest master.

lib/resources/r4/patient_patch.yaml
zsura: Updated cardinality to "0" and type to "N/A"
https://github.com/cerner/fhir.cerner.com/pull/242/commits/9a75bb6a3e406c75368166a12cb2d3fa039374ca

PR #250, opened 2019-04-04
Point all identifier.type references in terminology table to the documentation for identifier.type

krabey:


lib/resources/dstu2/soarian_device.yaml
```diff
@@ -9,13 +9,13 @@ fields:
   description: Identifier(s) from manufacturer, owner and others.
   children:
 
-  
+
```
MaxPhilips: can just take this line out rather than remove spaces from it

krabey: done: https://github.com/cerner/fhir.cerner.com/pull/250/commits/1c50d0d44d939152be7cb5a99f303209e8c89175#diff-fd3715a297194d5a88382f188ee54252L11

PR #266, opened 2019-05-31
Appointment R4 PATCH change 412 to 409

CheukwongZW:


content/millennium/r4/scheduling/appointment.md
```diff
@@ -29,7 +29,7 @@ When updating an appointment, the resource provides the ability to change the [A
 
 ## Patch
 
-Patch an existing appointment.
+Patch an existing Appointment.
```
CheukwongZW: made this change to be consistent with other resources

PR #274, opened 2019-07-18
Add Encounter Guarantor Documentation

beichhor:


content/millennium/dstu2/encounters/encounter.md
```diff
@@ -43,6 +43,19 @@ The following fields are returned if valued:
...
+
+* [Encounter guarantor]
+
+### Custom Extensions
+
+All URLs for custom extensions are defined as `https://fhir-ehr.cerner.com/dstu2/StructureDefinition/{id}`
+
+ ID                         | Value\[x] Type                                                          | Description
+----------------------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------
+ `encounter-guarantor`      | [`reference`](http://hl7.org/fhir/DSTU2/references.html) and [`period`] | The guarantors of an encounter. Contains a reference to a Patient, Person, or Organization and a period when the guarantor is effective.
```
MaxPhilips: since this is a complex extension, it feels like it should be three rows instead of one. since the nested extensions don't have their own structure definition ids, it feels like the first column cannot be titled id

let's start a new pattern for documenting these:
* on the StructureDefinition page
  * change the List of StructureDefinitions to have the full description that each resource page has now
  * move that list to a TOC-visible location on the page (probably after Overview and Retrieve by id; similar to Proprietary codes 'List of Code Sets')
* on each resource page
  * instead of organizing the table by StructureDefinition ID, organize it by extension

So in this case the table would be something like

 Extension | Value[x] Type | Description
---- | ---- | ----
Encounter Guarantor | None (contains nested extensions) | The guarantors of the encounter.
Encounter Guarantor Party | Reference | A reference to a Patient, Person, or Organization
Encounter Guarantor Period | Period | The period when this encounter guarantor is effective

beichhor: Still think we should include the ID of the StructureDefinition within the table or at-least on the resource page. Considering that the StructureDefinition page will link to the custom extensions section, it seems odd that some custom extension pages will contain the ID of the extension but others will not. Any objection to putting it in parentheses after the first row or creating a new column?

MaxPhilips: sure for rows of the table on the resource page that have StructureDefinition ids, adding them in parens sounds good

PR #289, opened 2019-09-11
Add Encounter.serviceType information to encounter and proprietary codes

JFisk42:


lib/resources/example_json/r4_examples_encounter.rb
```diff
@@ -253,6 +264,17 @@ module Resources
                 "text": "Inpatient"
               }
             ],
+            "serviceType": {
+              "coding": [
+                {
```
beichhor: Would this return back a standard code at any point? If so can we add it?

JFisk42: It would but mappings aren't live in fhirplay yet. I didn't want to add it before the mappings are there but I guess we probably should.

beichhor: Probably best to just get it out of the way now so we don't have to backfill it. I'll try and see what code would be best to use here.

JFisk42: Added: https://github.com/cerner/fhir.cerner.com/pull/289/commits/f7560d79a80c4100360333103a63244d9aacfe57

PR #292, opened 2019-09-17
Add Documentation for Condition retrieve by id

MadhurRaj:


lib/resources/example_json/r4_examples_condition.rb
```diff
@@ -104,6 +104,95 @@ module Resources
       ]
     }
 
+    R4_CONDITION_ENTRY ||= {
+      "fullUrl":"https://fhir-ehr.sandboxcerner.com/r4/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/Condition/p40880687",
```
beichhor: Is there any reason why there's no space after the `:`? The examples below do.

lib/resources/example_json/r4_examples_condition.rb
```diff
@@ -235,10 +324,10 @@ module Resources
         {
           "authorReference": {
             "reference": "Practitioner/1912007",
-            "display": "Test, Jack" 
+            "display": "Test, Jack"
           },
           "text": "Test Comment"
-          
+
```
beichhor: nit: remove extra space.

lib/resources/example_json/r4_examples_condition.rb
davischristi: yeah I guess we should be having a space after the colon.

PR #293, opened 2019-09-17
HTML escaping the narratives in the Encounter read and search examples

ebaranowski:
The Encounter Read and Search examples did not HTML escape the narrative in the JSON examples. It needed to be added so that the view rendered the HTML as code instead of rendering the HTML itself.

Validation:
Make sure the Encounter narratives are displaying HTML code.


beichhor:
Verified.
Screen Shot 2019-09-18 at 1 10 12 PM

PR #296, opened 2019-09-30
Add missing forward slash on Appointment PATCH example

beichhor:


beichhor:
Proof of change:

Screen Shot 2019-10-25 at 12 40 06 PM

PR #300, opened 2019-10-11
Change name and communication to array in R4_examples

vinaybhargavar:
The example for the R4 related person in the documentation shows name and communication as objects. But these two fields have a cardinality of 0..* in the fhir spec. This PR updates the documentation to accurately show the fields as an array.


sdesani:
Can we add the resource name, for which you are changing this, into the commit message ?

PR #308, opened 2019-10-25
kramdown warning when compiling site

MaxPhilips:
r4_examples_device contained some non-escaped < characters. escaped them in f4d8fee

R4 device.md contained some extra whitespace. removed it in 97a62a3


beichhor:
Can we put `Fixes #123` and then the PR number? It'll link them up and once this is merged will close the issue.

PR #314, opened 2019-10-31
Fix required fields documentation in R4 Coverage

cmccormick42:
Documentation on fhir.cerner.com indicates that Coverages can be created without supplying required fields subscriber or relationship, but both of these fields are required. The documentation now shows these fields as required with updated cardinality. 

Also, a typo in the period cardinality prevented it from being displayed on the page.

Local validation attached.

subscriber
relationship
period


lib/resources/r4/coverage.yaml
```diff
@@ -28,8 +28,8 @@ fields:
   note: Currently supports only `active`.
 
 - name: subscriber
-  required: 'No'
-  cardinality: 0..1
+  required: 'Yes'
+  cardinality: 1..1
```
MaxPhilips: the cardinality should still reflect the HL7 cardinality. required indicates that the field is Cerner-required

cmccormick42: Reverted in https://github.com/cerner/fhir.cerner.com/pull/314/commits/5f38d48b35ae270096f12ecb6da889ff84d9086c

cmccormick42:
Validation steps:
- Go to http://fhir.cerner.com/millennium/r4/financial/coverage/ 
- Scroll down to the Create section, under Body Fields. 
- Check that relationship shows Yes for required.
- Check that subscriber shows Yes for required. 
- Check that period shows 0..1 for cardinality.

vinaybhargavar:
Validated. I see that relationship and subscriber shows Yes for required and period shows 0..1 for cardinality.

Screen Shot 2019-11-07 at 10 51 25 AM

Screen Shot 2019-11-07 at 10 48 48 AM

Screen Shot 2019-11-07 at 10 50 04 AM

PR #315, opened 2019-10-31
Adding AllergyIntolerance terminology binding: RxNorm

CBrashear:
Adding RxNorm to the terminology bindings for r4. It's already documented in DSTU2, just updated in R4.

### DIFF
--- a/lib/resources/r4/allergy_intolerance.yaml
+++ b/lib/resources/r4/allergy_intolerance.yaml
@@ -98,6 +98,9 @@ fields:
                  or a negated/excluded code for a specific substance or class (e.g., "No latex allergy") or a general or
                  categorical negated statement (e.g., "No known allergy", "No known drug allergies").
     terminology:
+    - display: RxNorm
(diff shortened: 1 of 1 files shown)


lib/resources/r4/allergy_intolerance.yaml
```diff
@@ -98,6 +98,9 @@ fields:
                  or a negated/excluded code for a specific substance or class (e.g., "No latex allergy") or a general or
                  categorical negated statement (e.g., "No known allergy", "No known drug allergies").
     terminology:
+    - display: RxNorm
+      system: http://www.nlm.nih.gov/research/umls/rxnorm
+      info_link: http://hl7.org/fhir/dstu2/rxnorm.html
```
MaxPhilips: assuming this should be http://hl7.org/fhir/r4/rxnorm.html

CBrashear: Opps, fixed.

cmccormick42:
Validated: 
Screen Shot 2019-11-07 at 10 51 24 AM

PR #317, opened 2019-11-13
Overview Resources JSON Inconsistencies

wseymour15:
These changes only relate to resources in the Overview group. The goal of this PR is to let the user know that the JSON data from the example requests in a few of the examples are likely to change. Also, full paths were added to the conformance examples.

There will be more similar changes coming for the other resources (most likely much larger since there are some larger inconsistencies).

### DIFF
--- a/content/millennium/dstu2/conformance.md
+++ b/content/millennium/dstu2/conformance.md
@@ -37,22 +37,26 @@ Authorization is not required
 
 #### Request
 
-    GET /metadata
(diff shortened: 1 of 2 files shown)


wseymour15:
Verification:

Screen Shot 2019-11-21 at 9 55 19 AM
Screen Shot 2019-11-21 at 9 55 30 AM
Screen Shot 2019-11-21 at 9 55 44 AM

praneethk2094:
Live Verification:
Screen Shot 2019-11-25 at 4 22 03 PM
Screen Shot 2019-11-25 at 4 22 19 PM

PR #319, opened 2019-11-14
Devices Json Inconsistencies

wseymour15:
The goal of these changes were to add a disclaimer to request examples in the Device resource that the data may be different than the response shown.

Verifcation:
Screen Shot 2019-11-14 at 9 47 31 AM


praneethk2094:
Live Verification
Screen Shot 2019-11-25 at 4 24 23 PM

PR #320, opened 2019-11-14
Encounters JSON Inconsistencies

wseymour15:
The goal of these changes were to add a disclaimer to request examples in the Encounter resource that the data may be different than the response shown.

Verification:
Screen Shot 2019-11-14 at 9 56 40 AM


praneethk2094:
Live Verification: 
Screen Shot 2019-11-25 at 4 26 31 PM
Screen Shot 2019-11-25 at 4 26 11 PM

PR #333, opened 2019-12-17
Add Encounter External Indicator Documentation

sydneanschutz:
Terminologies:
Screen Shot 2019-12-17 at 2 41 07 PM
Screen Shot 2019-12-17 at 2 40 56 PM

Create: 
Screen Shot 2019-12-17 at 2 40 36 PM

### DIFF
--- a/lib/resources/r4/encounter.yaml
+++ b/lib/resources/r4/encounter.yaml
@@ -553,7 +553,7 @@ fields:
       }
 
 - name: meta.security
-  url: http://hl7.org/fhir/resource-definitions.html#Meta.security
+  url: https://hl7.org/fhir/r4/resource-definitions.html#Meta.security
   required: 'No'
   cardinality: 0..*
   type: Coding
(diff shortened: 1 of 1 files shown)


lib/resources/r4/encounter.yaml
```diff
@@ -169,6 +169,13 @@ fields:
     - display: Millennium Encounter Visit Type
       system: https://fhir.cerner.com/&lt;EHR source id&gt;/codeSet/71
       info_link: https://fhir.cerner.com/millennium/r4/proprietary-codes-and-systems/#code-set-71-visit-type
+    - display: v2 Patient Class
```
ebaranowski: I dont understand where this is coming from....

http://hl7.org/fhir/valueset-encounter-type.html

lib/resources/r4/encounter.yaml
```diff
@@ -544,3 +551,32 @@ fields:
       {
         "reference": "Location/35646787"
       }
+
+- name: meta.security
+  url: http://hl7.org/fhir/resource-definitions.html#Meta.security
```
ebaranowski: R4

lib/resources/r4/encounter.yaml
ebaranowski: Talked offline with Kathy. The value set is an example value set

lib/resources/r4/encounter.yaml
sydneanschutz: https://github.com/cerner/fhir.cerner.com/pull/333/commits/ecb183ebe47aae56128b75b511fccd7a46f190fc

ebaranowski:
Screen Shot 2019-12-17 at 3 03 55 PM
Screen Shot 2019-12-17 at 3 04 08 PM
Validated

averyallen:
image

PR #340, opened 2020-01-17
Set action to terminology for R4 Appointment reasonCode

### LINKED ISSUE #339: R4 Appointment documentation indicates reasonCode is available for creates
We missed scoping reasonCode in R4 Appointment to the terminology bindings table only. It should be present only there and not in the create body table.

MaxPhilips:
This change removes reasonCode from the R4 Appointment create body fields table. We did not set an action key in the R4 Appointment yaml file until now. R4 Appointment create ignores reasonCode if it is set.

Local verification - comment is now the final field in the create body fields table:
Screen Shot 2020-01-17 at 3 53 32 PM

Fixes #339

### DIFF
--- a/lib/resources/r4/appointment.yaml
+++ b/lib/resources/r4/appointment.yaml
@@ -175,6 +175,7 @@ fields:
   cardinality: 0..*
   type: CodeableConcept
   description: Coded reason this appointment is scheduled.
+  action: terminology
   binding:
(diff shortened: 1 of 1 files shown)


MaxPhilips:
waiting for next site deployment to capture post-merge verification

MaxPhilips:
post-deployment validation

Screen Shot 2020-01-29 at 3 45 51 PM

PR #348, opened 2020-02-14
Modifying identifier type and system examples to values that are supported

JFisk42:
Verification:

Screen Shot 2020-02-14 at 1 57 09 PM
Screen Shot 2020-02-14 at 1 57 22 PM
Screen Shot 2020-02-14 at 1 57 43 PM

### DIFF
--- a/lib/resources/example_json/r4_examples_encounter.rb
+++ b/lib/resources/example_json/r4_examples_encounter.rb
@@ -433,12 +433,13 @@ module Cerner
           "type": {
             "coding": [
               {
-                "system": "http://hl7.org/fhir/v2/0203",
-                "code": "FIN NBR"
+                "system": "https://fhir.cerner.com/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/codeSet/319",
+                "code": "1077",
+                "display": "FIN NBR"
               }
             ]
           },
-          "system": "urn:oid:192.168.0.1",
+          "system": "urn:oid:3.3.3.3.3.3",
           "value": "2000363999",
           "period": {
             "start": "2017-01-17T20:39:40.000Z"
--- a/lib/resources/example_json/r4_examples_encounter.rb
+++ b/lib/resources/example_json/r4_examples_encounter.rb
@@ -427,25 +427,6 @@ module Cerner
     R4_ENCOUNTER_CREATE ||= {
       "resourceType": "Encounter",
       "status": "in-progress",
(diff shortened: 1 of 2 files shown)


lib/resources/example_json/r4_examples_encounter.rb
```diff
@@ -433,12 +433,13 @@ module Resources
           "type": {
             "coding": [
               {
-                "system": "http://hl7.org/fhir/v2/0203",
-                "code": "FIN NBR"
+                "system": "https://fhir.cerner.com/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/codeSet/319",
```
MadhurRaj: I'd prefer not to have this (writing a FIN) listed in the actual example. I think it's fine to have it in the body fields to explain what you could do with the field, but not list it out in the example.

My concern is that FINs should mostly be system generated. This example makes it seem like we're encouraging adding a FIN manually.

JFisk42: That's a good point. I can try to find a better example for identifier type that works in sandbox because I still think the purpose of these examples should be to show what a maximal example body looks like.

JFisk42: How would you feel about using Visit Id?

`            "type": {
                "coding": [
                    {
                        "system": "https://fhir.cerner.com/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/codeSet/319",
                        "code": "1081",
                        "display": "Visit Id"
                    }
                ],
                "text": "Visit Id"
            },`

Side note: I am also purposefully including the display even though on a write it isn't considered. Seems appropriate for a proprietary code.

MadhurRaj: Seems like Visit Id could also be generated. I'm leaning back towards my original suggestion to not include it in the example.

If we are, I guess we can just stick with FIN.

FFSol:
@MadhurRaj stripped the generated identifier example after failing to locate an agreed-upon non-generated example.

FFSol:
Screen Shot 2020-04-03 at 9 15 50 AM
Stripped identifier verification (locally)

ebaranowski:
Validated:

Screen Shot 2020-04-07 at 5 15 55 PM
Screen Shot 2020-04-07 at 5 16 10 PM
Screen Shot 2020-04-07 at 5 16 32 PM

PR #352, opened 2020-03-03
Observation formatted comment documentation update

BS062721:
DSTU2:
DSTU2_Comment_information

R4:
R4_comment_information

### DIFF
--- a/content/millennium/r4/diagnostic/observation.md
+++ b/content/millennium/r4/diagnostic/observation.md
@@ -55,7 +55,7 @@ Search for labs, vitals, and alcohol/tobacco use Observations that meet supplied
 
 _Implementation Notes_
 
-* The `comments` field may have RTF or other formatted data rather than simple text. We are investigating markdown support.
+* The `comments` field may have RTF or other formatted data rather than simple text. This is an issue that will be resolved in a future correction. We are also investigating markdown support.
 
 * When multiple pages of Observation results are returned for a single query:
   * All Social history Observations (if any qualify for the query) will be returned on the first page of results. This means that the next bullet does not apply to Social history Observations.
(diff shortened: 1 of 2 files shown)


content/millennium/r4/diagnostic/observation.md
```diff
@@ -55,7 +55,7 @@ Search for labs, vitals, and alcohol/tobacco use Observations that meet supplied
 
 _Implementation Notes_
 
-* The `comments` field may have RTF or other formatted data rather than simple text.
+* The `comments` field may have RTF or other formatted data rather than simple text. This is an issue that will be resolved in a future correction. We are also investigating markdown support.
```
beichhor: I think this note may be still too vague.

BS062721: This was suggested by Dennis, I am not involved in that conversation Seada can give more information on this.

SA030800: yes this is the note from dennis

> For DSTU2, we already state there is RTF.  We’d update the existing comment “This is an issue that will be resolved in a future correction” to instead say it’ll be resolved in R4.  For R4, we’d echo the existing comment about RTF / formatted data and state that we’re investigating markdown support.

BS062721: Updated it as discussed: https://github.com/cerner/fhir.cerner.com/pull/352/commits/56249e811fbbf88e70affc672b4f98dae6a85ef6

New output:

Screen Shot 2020-03-04 at 12 07 38 PM

ebaranowski:
Screen Shot 2020-04-01 at 10 23 26 AM
Screen Shot 2020-04-01 at 10 23 55 AM

Validated

PR #353, opened 2020-03-09
Add Documentation for MedicationRequest creates

beichhor:
Description
----

PR Checklist
----

Local verification:


content/millennium/r4/medications/medication-request.md
```diff
@@ -139,10 +139,54 @@ We are currently always returning a value of "Order" in the intent field. We rec
...
+Last-Modified: Wed, 27 Mar 2019 15:59:30 GMT
+Location: https://fhir-ehr.sandboxcerner.com/r4/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/MedicationRequest/4595905
+Server-Response-Time: 3890.363996
+Status: 201 Created
+Vary: Origin
+X-Request-Id: 1638e30e497b93ff4383b2ff0eaeea68
+X-Runtime: 3.890282
+</pre>
+
+The `ETag` response header indicates the current `If-Match` version to use on subsequent updates.
```
MaxPhilips, line 188: probably leave this note off of resources that don't expose updates?

content/millennium/r4/medications/medication-request.md
```diff
@@ -139,10 +139,54 @@ We are currently always returning a value of "Order" in the intent field. We rec
...
+
+### Errors
+
+The common [errors] and [OperationOutcomes] may be returned.
```
MaxPhilips, line 192: the markdown links at the end of this file got deleted

lib/resources/example_json/r4_examples_medication_request.rb
```diff
@@ -438,79 +438,104 @@ module Resources
...
+        "text": "Acetaminophen"
+      },
+      "subject": {
+        "reference": "Patient/1316024"
+      },
+      "encounter": {
+        "reference": "Encounter/1621910"
+      },
+      "note": [
+    		{
```
MaxPhilips, line 467: convert all tab characters in this file to spaces

lib/resources/r4/medication_request.yaml
```diff
@@ -16,6 +34,7 @@ fields:
   required: 'No'
   type: CodeableConcept
   description: Reason for the current status
+  action: get
```
MaxPhilips, line 37: don't think get is a supported action for tables

lib/resources/r4/medication_request.yaml
```diff
@@ -2,10 +2,28 @@
...
 - name: status
   required: 'Yes'
   type: code
   description: The current state of the medication request.
+  action: create
+  example: |
+    {
+      "status": "active"
+    }
+  note: Currently we only the support the `active` status.
```
MaxPhilips, line 26: rewrite all 'currently we only support the x field' as 

```
When creating a new MedicationRequest, the status must be `active`
```

content/millennium/r4/medications/medication-request.md
HunterBerten: Removed.

content/millennium/r4/medications/medication-request.md
HunterBerten: Readded.

lib/resources/example_json/r4_examples_medication_request.rb
HunterBerten: Fixed.

lib/resources/r4/medication_request.yaml
HunterBerten: Should it be removed then, or should it be another action? Or do we need confirmation on it?

lib/resources/r4/medication_request.yaml
HunterBerten: Reworded.

content/millennium/r4/medications/medication-request.md
```diff
@@ -139,6 +139,56 @@ We are currently always returning a value of "Order" in the intent field. We rec
...
+<pre class="terminal">
+Cache-Control: no-cache
+Content-Length: 0
+Content-Type: application/fhir+json
+Date: Wed, 27 Mar 2019 15:59:33 GMT
+Etag: W/"0"
+Last-Modified: Wed, 27 Mar 2019 15:59:30 GMT
+Location: https://fhir-ehr.sandboxcerner.com/r4/0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca/MedicationRequest/4595905
+Server-Response-Time: 3890.363996
+Status: 201 Created
```
daliboz, line 182: Please remove Status - I thought we already removed this non-standard from the response, but it causes some confusion. May consider removing others that we aren't purposefully adding for devs to use? (eg: X-Runtime and Server-Response Time?)

lib/resources/r4/medication_request.yaml
```diff
@@ -48,26 +79,88 @@ fields:
...
   type: code
   description: Medication Request Priority
+  action: get
   binding:
     description: Medication Request Priority
     terminology:
     - display: Request Priority
       system: http://hl7.org/fhir/request-priority
       info_link: https://hl7.org/fhir/r4/valueset-request-priority.html
+- name: doNotPerform
```
daliboz, line 89: What does Millennium do with this field?

lib/resources/r4/medication_request.yaml
```diff
@@ -48,26 +79,88 @@ fields:
...
+    {
+      "doNotPerform": false
+    }
+- name: reported[x]
+  required: 'Yes'
+  cardinality: 0..1
+  type: boolean
+  description: Indicates if this record was captured as a secondary 'reported' record rather than as an original primary source-of-truth record.
+  action: create
+  note: Must be of type boolean and must be set to `true`.
```
daliboz, line 106: Add a similar statement to read/search: The server may change the stored value following business rules and pending changes to regulation. Needs better language.

content/millennium/r4/medications/medication-request.md
HunterBerten: Removed.

lib/resources/r4/medication_request.yaml
HunterBerten: Sorry, where exactly should this be added in read/search? Just like in the implementation notes section? Also not really sure how it should be worded. @kathypickering?

lib/resources/r4/medication_request.yaml
daliboz: According to that wiki, we also don't support the field (it's greyed out). If that's true, why are we documenting it?

lib/resources/r4/medication_request.yaml
daliboz: You're adding a statement in create similar to what we added to read/search about the intent and changes pending regulatory. I believe we currently will accept either value, but it will be read back out as "order" - but in the future we will change that to "plan" following US Core guidelines

content/millennium/r4/medications/medication-request.md
daliboz: Thoughts about the other fields I mentioned?

HunterBerten: I'm not really sure. Personally, I'd say if it's not useful it's not really worth clogging up the page, but all that is listed on the other resources and I think it should at least be consistent.

lib/resources/r4/medication_request.yaml
HunterBerten: Looking through the code, it seems like we do support it. I think the wiki probably just isn't up to date? Evan had to update it while he was removing the JBuilders, so I think that's the case for at least a few resources.

content/millennium/r4/medications/medication-request.md
daliboz: We caused some confusion when status was removed. Technically, these are created when you make the request (status wouldn't be in the current response). I don't want to document headers in these examples that we're not intended intentionally to produce, in order to avoid the issue we have with status (Status is also listed in other examples, which are now out of date). I always will vote for "document with intent" rather than accident :)

lib/resources/r4/medication_request.yaml
HunterBerten: Got it- wasn't sure since the comment was on the reported field. Is keeping it the same as the reach/search version alright? Possibly specifying that we accept order and plan for creates, like:

**Implementation Notes**

We accept "Order" and "Plan" in the intent field, but are currently always returning a value of "Order." We recommend using the reported field to determine if a medication is an authorization or a medication reported by a patient. In the future, we will support "Plan" and "Order" in accordance with the US Core Profile pending changes.

content/millennium/r4/medications/medication-request.md
HunterBerten: I'd say X-Runtime and Server-Response-Time probably should be removed then, but I'm not really 100% which ones we're adding for devs to use 😅

lib/resources/r4/medication_request.yaml
MaxPhilips: well i'm pretty sure it doesn't do anything differently present or absent - you can confirm with some testing locally

lib/resources/r4/medication_request.yaml
MaxPhilips: status->intent in the second one

content/millennium/r4/medications/medication-request.md
MaxPhilips: fine to remove for this PR and we will loop through other examples later https://github.com/cerner/fhir.cerner.com/issues/397

lib/resources/r4/medication_request.yaml
HunterBerten: Fixed: https://github.com/cerner/fhir.cerner.com/pull/353/commits/5b4b8c951640267e7d7ac2be448cdeae49b511d0

content/millennium/r4/medications/medication-request.md
HunterBerten: Removed X-Runtime and Server-Response-Time: https://github.com/cerner/fhir.cerner.com/pull/353/commits/5b4b8c951640267e7d7ac2be448cdeae49b511d0 Let me know if you two think others should be removed as well, since like I said I'm not 100% sure.

lib/resources/r4/medication_request.yaml
MaxPhilips: it seems to me that this field is validated but unused, we can loop back on how that became the case. however, since it is validated, i think we should keep this note

MaxPhilips: millennium doesn't use this field but since it is a modifier we verify that it cannot be true

content/millennium/r4/medications/medication-request.md
MaxPhilips: removing just those headers is ok

lib/resources/r4/medication_request.yaml
MaxPhilips: looks like these are being used to keep fields out of the create table

lib/resources/r4/medication_request.yaml
MaxPhilips: https://github.com/cerner/fhir.cerner.com/pull/353/commits/387ed3fddb74d45d41d8b7765a48b4f7b87ce43e#diff-152ba56eff926bdcdfebbc13bbde483fR57

DanielWhiteCerner:
screencapture-fhir-cerner-millennium-r4-medications-medication-request-2020-06-16-16_22_35
screencapture-fhir-cerner-millennium-r4-medications-medication-request-2020-06-16-16_22_35-2
screencapture-fhir-cerner-millennium-r4-medications-medication-request-2020-06-16-16_22_35-3

PR #354, opened 2020-03-18
Update General Clinical JSON inconsistencies

adamb3ard:
I went through the General Clinical section and fixed some inconsistencies associated with their JSON. (My IDE also removed some trailing white space.)

**AllergyIntolerance**:
I changed the UPDATE example to use the closed endpoint and updated the URL to match the ID in the JSON.
Closed endpoint:
AlgInUpClosedEP
Matching Ids:
AlgInUpMatchingID

**Condition**:
I updated the Health-Concern examples to use one that currently exists. This was done so the retrieve by id example will not return a 404. The Create and Update examples have been given a full, closed URL, and the Update example has been modified to an example that should succeed.
Updated read with a new Health-Concern:
ConNewReadEx
Condition Update change:
ConUpdate

**AllergyIntolerance, Condition, & Procedure**:
I added the disclaimer message after each JSON example in these 3 resources.
disclaimer


MaxPhilips:
can you fix the typo in the commit message?

adamb3ard:
@MaxPhilips Updated

ebaranowski:
Screen Shot 2020-04-01 at 10 26 25 AM
Screen Shot 2020-04-01 at 10 26 59 AM
Screen Shot 2020-04-01 at 10 27 53 AM
Screen Shot 2020-04-01 at 10 28 11 AM
Screen Shot 2020-04-01 at 10 28 36 AM
Screen Shot 2020-04-01 at 10 29 05 AM

Validated

PR #355, opened 2020-03-20
Adding notes about video visits

rlcern:
Adding notes about video visit functionality, since the documentation does not reflect that it is for specific installations.


ebaranowski:
Screen Shot 2020-04-01 at 10 32 00 AM

Validated

PR #358, opened 2020-03-25
Add example System Account token request

joeratt:
Proposing an example cURL to highlight for developers to create a system token.

I'm adding a second scope to the `scopes` example, just to highlight to developers that scopes are space-delimited.


joeratt:
Screen Shot 2020-03-25 at 13 03 00

Attaching evidence of how this change looks locally

ebaranowski:
Screen Shot 2020-04-01 at 10 36 52 AM

Validated

PR #360, opened 2020-03-26
Update Medications JSON inconsistencies

adamb3ard:
I went through the DSTU2 Medications section and fixed some inconsistencies with the JSON.

**MedicationStatement**:
The current create example does not work so I updated the Patient to one that exists, so that it does.
EX:
medCreate

The Update throws a 404, so I updated it to use a currently existing MedicationStatement and Patient. This call will not succeed since it has already been updated to complete, but now that it is tied to an existing MedicationStatement a user can read that resource to see its status.
EX:
medUpdate

**All**:
I also went through and added the disclaimer to all the Medications response examples.
EX:
medDisclaimer

### DIFF
--- a/content/millennium/dstu2/medications/immunization.md
+++ b/content/millennium/dstu2/medications/immunization.md
@@ -74,6 +74,8 @@ Notes:
 <%= headers status: 200 %>
 <%= json(:dstu2_immunization_bundle) %>
 
+<%= disclaimer %>
+
 ### Errors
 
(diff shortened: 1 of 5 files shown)


ebaranowski:
Screen Shot 2020-04-01 at 10 34 58 AM
Screen Shot 2020-04-01 at 10 35 15 AM
Screen Shot 2020-04-01 at 10 35 28 AM

Validated

PR #365, opened 2020-03-31
Add documentation for R4 Condition diagnosis update

sydneanschutz:
Screen Shot 2020-03-31 at 9 02 34 AM
Screen Shot 2020-03-31 at 9 03 42 AM

### DIFF
--- a/content/millennium/r4/general-clinical/condition.md
+++ b/content/millennium/r4/general-clinical/condition.md
@@ -178,8 +178,10 @@ Update an existing condition.
 
 _Implementation Notes_
 
-* Currently `problem-list-item` is supported.
+* Currently `problem-list-item` and `encounter-diagnosis` are supported.
 * Any field which is missing will be interpreted as nulling out or removing data from the resource. See [FHIR<sup>®</sup> Update] for additional details about update operations.
+* `abatementDateTime` is not supported for Conditions with a category of `encounter-diagnosis`.
+* `note` cannot be added if note already exists. An existing note field cannot be modified.
 
 ### Authorization Types
 
--- a/content/millennium/r4/general-clinical/condition.md
+++ b/content/millennium/r4/general-clinical/condition.md
@@ -178,7 +178,7 @@ Update an existing condition.
 
 _Implementation Notes_
 
-* Currently `problem-list-item` and `encounter-diagnosis` are supported.
+* Currently `problem-list-item` is supported.
 * Any field which is missing will be interpreted as nulling out or removing data from the resource. See [FHIR<sup>®</sup> Update] for additional details about update operations.
 
 ### Authorization Types
--- a/content/millennium/r4/general-clinical/condition.md
+++ b/content/millennium/r4/general-clinical/condition.md
@@ -180,8 +180,6 @@ _Implementation Notes_
 
 * Currently `problem-list-item` and `encounter-diagnosis` are supported.
 * Any field which is missing will be interpreted as nulling out or removing data from the resource. See [FHIR<sup>®</sup> Update] for additional details about update operations.
-* `abatementDateTime` is not supported for Conditions with a category of `encounter-diagnosis`.
-* `note` cannot be added if note already exists. An existing note field cannot be modified.
 
 ### Authorization Types
 
--- a/content/millennium/r4/general-clinical/condition.md
+++ b/content/millennium/r4/general-clinical/condition.md
@@ -178,7 +178,7 @@ Update an existing condition.
 
 _Implementation Notes_
 
-* Currently `problem-list-item` is supported.
+* Currently `problem-list-item` and `encounter-diagnosis` are supported.
 * Any field which is missing will be interpreted as nulling out or removing data from the resource. See [FHIR<sup>®</sup> Update] for additional details about update operations.
 
 ### Authorization Types
--- a/lib/resources/r4/condition.yaml
+++ b/lib/resources/r4/condition.yaml
@@ -335,8 +335,7 @@ fields:
     <ul>
       <li> Only a type of dateTime is supported. </li>
       <li> The dateTime is converted to a GMT date. </li>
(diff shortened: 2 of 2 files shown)


content/millennium/r4/general-clinical/condition.md
```diff
@@ -178,8 +178,10 @@ Update an existing condition.
 
 _Implementation Notes_
 
-* Currently `problem-list-item` is supported.
+* Currently `problem-list-item` and `encounter-diagnosis` are supported.
 * Any field which is missing will be interpreted as nulling out or removing data from the resource. See [FHIR<sup>®</sup> Update] for additional details about update operations.
+* `abatementDateTime` is not supported for Conditions with a category of `encounter-diagnosis`.
+* `note` cannot be added if note already exists. An existing note field cannot be modified.
```
cmccormick42, line 184: Is this true of problems as well?

sanschutz0808: This is actually a bit confusing because on our mappings wiki it says "For problems there is nothing in local service to prevent updates" but [here](https://github.com/cerner/fhir.cerner.com/blob/master/lib/resources/r4/condition.yaml#L336) this information was already published. Under condition updates:
Screen Shot 2020-04-03 at 9 25 42 AM

cmccormick42:
> I think we should add the encounter field documentation as well. Even though it's optional since local service can't use it, it still gets validated if it's provided.

Thought some more about this, probably doesn't make sense to advertise that field if it can't be updated.

content/millennium/r4/general-clinical/condition.md
nishanthnamana: So for problems, if a particular condition doesn't have a note and the user is sending a note in the update request body, in that case we write note to that particular condition. But if there is already a note available for a condition, then local service will ignore the note from the update request body.

nishanthnamana, line 184: I prefer adding them at the field level.

sanschutz0808: Removed: https://github.com/cerner/fhir.cerner.com/pull/365/commits/2eb7b4a52698b32cdb3c8d3f9d0af6a11cbafb22#diff-c91ea4243fc73689536dc746c972a3d6L183

sanschutz0808: Kept note here: https://github.com/cerner/fhir.cerner.com/pull/365/commits/2eb7b4a52698b32cdb3c8d3f9d0af6a11cbafb22#diff-abb99d2139d79f8a2a8f1d214f5f6882R363

Is this correct?

sanschutz0808:
> Aren't we missing encounter field?
> Also we need to update notes for 'abatementDateTime' saying it is only support when category is 'problem-list-item'.
> We also need to update notes for clinical-status and verification-status, since we only support few values. We need to add our 422 scenario's as notes for these fields.

Added encounter field (confirmed with Fran): https://github.com/cerner/fhir.cerner.com/pull/365/commits/2eb7b4a52698b32cdb3c8d3f9d0af6a11cbafb22#diff-abb99d2139d79f8a2a8f1d214f5f6882R271

abatementDateTime update: https://github.com/cerner/fhir.cerner.com/pull/365/commits/2eb7b4a52698b32cdb3c8d3f9d0af6a11cbafb22#diff-abb99d2139d79f8a2a8f1d214f5f6882R316

clinical-status: https://github.com/cerner/fhir.cerner.com/pull/365/commits/2eb7b4a52698b32cdb3c8d3f9d0af6a11cbafb22#diff-abb99d2139d79f8a2a8f1d214f5f6882R73

I think we still support all verification-statuses?

lib/resources/r4/condition.yaml
```diff
@@ -263,6 +268,26 @@ fields:
...
+  example: |
+    {
+      "encounter": {
+        "reference": "Encounter/631923"
+      }
+    }
+  note: |
+    <ul>
+      <li> encounter must be a reference to an Encounter. </li>
+      <li> encounter during which the condition was first asserted. </li>
```
cmccormick42, line 285: Might not need this here since it's already in the description.

Should we mention that the encounter can't actually be updated?

sanschutz0808: Removed: https://github.com/cerner/fhir.cerner.com/pull/365/commits/ad905dea3880b346653ab91e60519e03aecd5bbf

They don't mention it in dstu2 and I was going off of that. https://fhir.cerner.com/millennium/dstu2/general-clinical/condition/#update

sanschutz0808:
> I don't think we need to add this, we are specifying this in the description.

addressed above

> when verificationStatus is 'refuted'?

or refuted: https://github.com/cerner/fhir.cerner.com/pull/365/commits/fb4343890ed723236f4c40d84fd64316a6accf84

nishanthnamana:
> > I don't think we need to add this, we are specifying this in the description.
> 
> addressed above
> 
> > when verificationStatus is 'refuted'?
> 
> or refuted: [fb43438](https://github.com/cerner/fhir.cerner.com/commit/fb4343890ed723236f4c40d84fd64316a6accf84)

+1

content/millennium/r4/general-clinical/condition.md
nishanthnamana: +1

nishanthnamana:
**Local Validation for Updates**
Update
diagnosis_update

ClinicalStatus
diagnosis_update_clinicalStatus

Category
diagnosis_update_category

Encounter
diagnosis_update_encounter

AbatementDateTime
diagnosis_update_abatementDateTime

nishanthnamana:
+1 after resolving conflicts.

MaxPhilips:
Screen Shot 2020-04-07 at 5 21 28 PM
Screen Shot 2020-04-07 at 5 21 43 PM
validated deployment

PR #367, opened 2020-04-01
Update Entities JSON inconsistencies

adamb3ard:
I went through the Entities resources in DSTU2 and updated some JSON inconsistencies.

Person:
I updated the combined person entry to one that currently exists.
personLinkUpdate

In Person and Contract:
I went through and added the disclaimer after each response. A working example of this can also be seen in the previous picture.


MaxPhilips:
Screen Shot 2020-04-07 at 5 22 57 PM
Screen Shot 2020-04-07 at 5 23 42 PM
validated deployment, one issue

MaxPhilips:
for person we added the disclaimer to the read by id example but not the search example

PR #370, opened 2020-04-03
Add redirect_uri requirement to Authorization documentation

shriniketsarkar:
Updated the authorization documentation to include Cerner's requirement of the redirect_uri to be passed along with other parameters to the call for authorization.


content/authorization.md
```diff
@@ -382,6 +383,15 @@ where the audience is provided:
 &aud=https%3A%2F%2Ffhir-ehr.sandboxcerner.com%2Fdstu2%2F0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca%2F
 </pre>
 
+#### Redirect URI ####
+
+When performaing an authorization grant request Cerner requires the application to
```
icon5585, line 388: It is not required when requesting the authorization code, it is optional, if not provided will use registered one

content/authorization.md
```diff
@@ -382,6 +383,15 @@ where the audience is provided:
...
 </pre>
 
+#### Redirect URI ####
+
+When performaing an authorization grant request Cerner requires the application to 
+provide with the redirect uri. This uri should match with the redirect uri setup during 
+the registration phase of the SMART Application. 
+
+The redirect uri must be an absolute uri. For more information please consult the 
+RFC 6749 [Redirection Endpoint Ref](https://tools.ietf.org/html/rfc6749#section-3.1.2)
```
icon5585, line 396: Otherwise looks good

content/authorization.md
daliboz: +1 - if they have multiple redirects, they should specify, or it will use the primary one configured.

icon5585:
+1

content/authorization.md
icon5585: @daliboz  - That is correct, we will redirect to the primary redirect URI, not the additional redirect URIs that are registered.

shriniketsarkar: Addressed : https://github.com/cerner/fhir.cerner.com/pull/370/commits/85110eda39d7034a9be61302a9c2251f700a489c

icon5585:
Did we lose the part around the requirement that the redirect URI be an absolute URI?

shriniketsarkar:
@icon5585 Its mentioned here :  https://github.com/cerner/fhir.cerner.com/pull/370/files#diff-1f4e364b44a831b80cc3c8de3ea1cd07R397-R398

content/authorization.md
```diff
@@ -382,6 +383,20 @@ where the audience is provided:
 &aud=https%3A%2F%2Ffhir-ehr.sandboxcerner.com%2Fdstu2%2F0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca%2F
 </pre>
 
+#### Redirect URI ####
+
+When performaing an authorization grant request the application may provide a redirect uri.
```
kolkheang, line 388: performaing -> performing

We should make it clear that in situation where there are multiple registered redirect URIs, they must provide a `redirect_uri`.

content/authorization.md
```diff
@@ -382,6 +383,20 @@ where the audience is provided:
 &aud=https%3A%2F%2Ffhir-ehr.sandboxcerner.com%2Fdstu2%2F0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca%2F
 </pre>
 
+#### Redirect URI ####
+
+When performaing an authorization grant request the application may provide a redirect uri. 
+This uri should match the redirect uri setup during the registration phase of the SMART Application.
+
+If your application has multiple redirect uri's registered with cerner then a redirect uri is expected to be passed
```
kolkheang, line 391: cerner -> Cerner

content/authorization.md
```diff
@@ -382,6 +383,20 @@ where the audience is provided:
...
+in during the authorization grant request. If no redirect uri is specified the default redirect uri 
+registered for the application will be used.
+
+Details around supplying a redirect uri are specified in the [SMART Specification](http://www.hl7.org/fhir/smart-app-launch/#step-1-app-asks-for-authorization)
+
```
kolkheang, line 396: Please add an inline example code block to show the construction of the `redirect_uri` segment.

content/authorization.md
```diff
@@ -382,6 +383,20 @@ where the audience is provided:
 &aud=https%3A%2F%2Ffhir-ehr.sandboxcerner.com%2Fdstu2%2F0b8a0111-e8e6-4c26-a91c-5069cbc6b1ca%2F
 </pre>
 
+#### Redirect URI ####
+
```
kolkheang, line 387: We should add an error example for this in the troubleshooting section so people know how to fix their issue with multiple redirect URIs situation.

content/authorization.md
shriniketsarkar: @kolkheang  Wouldnt using  ```must``` imply that its required for the multiple redirect uri scenarios ?
Since its optional hence I choose ```may provide a redirect_uri```

content/authorization.md
shriniketsarkar: Can you elaborate on which error you are referring to ?

content/authorization.md
shriniketsarkar: Added an inline example block and a complete example link : https://github.com/cerner/fhir.cerner.com/pull/370/commits/f833a4f6ab9a21f7028085efb35ab19ef0a578b7

content/authorization.md
shriniketsarkar: Fixed the typo here : https://github.com/cerner/fhir.cerner.com/pull/370/commits/f833a4f6ab9a21f7028085efb35ab19ef0a578b7

Still waiting on thoughts around using ```must``` as compared to ```may```.

content/authorization.md
shriniketsarkar: Fixed : https://github.com/cerner/fhir.cerner.com/pull/370/commits/f833a4f6ab9a21f7028085efb35ab19ef0a578b7

content/authorization.md
kolkheang: Sorry for the delay!  I just saw this additional comment.

I was referring to the situation where there are multiple registered redirect URIs but the application isn't sending a redirect URI during the authorize flow.  I assume that's the error that the user is seeing today which led to this update in the documentation.

In this case, what error does the user see?

content/authorization.md
kolkheang: I just re-read the wording in this section and what you have is fine with regard to the wording of may vs must.

👍

content/authorization.md
daliboz: I thought this defaulted to the first listed/primary. @icon5585 - is that what we confirmed?

icon5585: @daliboz that is correct, if no redirect uri is provided then we default to the primary redirect uri that is registered

shriniketsarkar: @kolkheang  I think the error that the user saw was because, the redirect URL was not an absolute URL. More of the type : ```abc://``` instead of ```abc://callback```. 

Considering the above I am not adding an error scenario.

MaxPhilips:
local verification
Screen Shot 2020-05-22 at 1 27 05 PM

BS062721:
Verification after publish:

Screen Shot 2020-05-22 at 1 41 07 PM

PR #379, opened 2020-04-30
Fix Encounter R4 examples

beichhor:
Description
----
Adds the Accept header for the Create example and switches the participant.period and participant.individual examples.
Screen Shot 2020-04-30 at 3 58 48 PM
Screen Shot 2020-04-30 at 3 57 53 PM

PR Checklist
----

### DIFF
--- a/content/millennium/r4/encounters/encounter.md
+++ b/content/millennium/r4/encounters/encounter.md
@@ -136,7 +136,7 @@ Create an individual Encounter.
 
 ### Headers
 
-<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', 'Content-Type': 'application/fhir+json'} %>
+<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', 'Accept': 'application/fhir+json', 'Content-Type': 'application/fhir+json'} %>
 
 ### Body Fields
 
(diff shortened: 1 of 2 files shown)


content/millennium/r4/encounters/encounter.md
```diff
@@ -136,7 +136,7 @@ Create an individual Encounter.
 
 ### Headers
 
-<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', 'Content-Type': 'application/fhir+json'} %>
+<%= headers head: {Authorization: '&lt;OAuth2 Bearer Token>', 'Accept': 'application/fhir+json', 'Content-Type': 'application/fhir+json'} %>
```
ebaranowski, line 139: Why is there an Accept header on a Create?

beichhor: We're relatively inconsistent in our exmaples for creates https://fhir.cerner.com/millennium/r4/individuals/patient/#create we either need to add this here or remove it from other places.

ebaranowski: Hmmmm. So best practice, I would think it would be to have an Accept header. The reason you dont need one is because we default to json. So if we have it in other places, then I think it's fine to leave it.

shriniketsarkar:
Validated here : https://fhir.cerner.com/millennium/r4/encounters/encounter/#create
Screen Shot 2020-05-22 at 2 27 18 PM
Screen Shot 2020-05-22 at 2 27 46 PM

PR #381, opened 2020-05-12
Add R4 Encounter sorting details, update Pagination with previous link

FFSol:
Description
----

- Document that R4 Encounter Bundles are sorted by the Encounter.period
- Update Bundle Page links to include the Previous link

PR Checklist
----
- [x] Screenshot(s) of changes attached before changes merged.
- [ ] Screenshot(s) of changes attached after changes merged and published.

PR #398, opened 2020-06-18
Further clarify R4 Encounter sorting details

FFSol:
Description
----
Addressing slightly ambiguous verbiage in #381 

PR Checklist
----
- [ ] Screenshot(s) of changes attached before changes merged.


FFSol:
Screen Shot 2020-06-18 at 11 03 32 AM

beichhor:
I don't think this wording really makes it clearer. We need to state that the Encounters returned per page will be more recent than Encounters on subsequent pages. We need to also be clear that within a page the sorting is not guaranteed.

FFSol:
Updated screenshot:
Screen Shot 2020-07-23 at 10 57 40 AM

MaxPhilips (approved):
+1 if we ditch the second period at the end

PR #419, opened 2020-07-08
Add RSpec tests, rubocop, and GitHub Workflows

beichhor:
Description
----
This adds GitHub actions workflows that will build against the master branch and open PRs. Rspec tests are ran and rubocop lint is checked for. This PR includes set up for RSpec, tests for the resources.rb file (additional tests will be added later) and resolutions for existing lint. 

There are no content changes related to this change.

PR Checklist
----
- [x] ~Screenshot(s) of changes attached before changes merged.~


.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    - name: Checkout repo.
+      uses: actions/checkout@v2
+
+    - name: Setup Ruby
+      uses: ruby/setup-ruby@v1.38.0
```
sdesani, line 24: Just curious, how we landed on using `ruby/setup-ruby` vs `actions/setup-ruby` ? I could not find more details on differences between two.

.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Install Gems
+      run: bundle install
+
+    - name: Run tests
```
sdesani, line 29: nit: `Run Tests`

.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: rspec
+
+    - name: Run rubocop
```
sdesani, line 32: nit: `Rub Rubocop`

.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    - name: Checkout repo.
```
sdesani, line 20: nit: We can remove period after repo.

.github/workflows/main.yml
sdesani, line 24: We can also probably add a comment here suggesting what version it installs
Something like - `# Installs version mentioned in .ruby-version file`

sdesani:
Might want to rename the yaml file from main.yml to may be something like `continuous-integration-workflow.yml`. We might end up adding more workflows (like publishing the site) later so to keep it unique its better to have a specific names.

.github/workflows/main.yml
beichhor: This is based off of the template GitHub generated. That is why I had used ruby/setup-ruby

.rubocop.yml
```diff
@@ -0,0 +1,258 @@
+AllCops:
+  TargetRubyVersion: 2.6
+  Exclude:
+    - 'lib/resources/example_json/*.rb'
```
MaxPhilips, line 4: let's lint the example json files too

beichhor: I want to be clear, if we actually lint the example JSON files this will require us to fix almost an additional 10,000 lint errors. (that's not an exaggeration). This is _primarily_ because of the `Style/StringLiterals` linter (preferring single quotes strings), but there's also the `Layout/LineLength` style as well. My primary reason for disabling this was because of the amount of changes and we mostly copy and paste the examples from sandbox and don't alter them in any way. If you want me to I can enable it, but the changes this will introduce will certainly outweigh the rest of this PR. 

It might be better if the lint errors specifically for the examples are updated and resolved (with the disabling removed) in this PR: https://github.com/cerner/fhir.cerner.com/pull/420 since that is updating the examples already.

beichhor: Logged issue: https://github.com/cerner/fhir.cerner.com/issues/423

beichhor:
This shows the diff of the site from this branch output and the gh-pages branch. The only changes are the ones introduced in previous commits.

Screen Shot 2020-07-10 at 10 46 15 AM

daliboz:
@beichhor You will need to go through and change the branch referenced to main :)

.rubocop.yml
```diff
@@ -115,20 +119,14 @@ Metrics/ClassLength:
...
-Metrics/LineLength:
-  Max: 120
-
 # Avoid methods longer than 30 lines of code
 Metrics/MethodLength:
-  Exclude:
-    - 'engines/*/spec/**/*'
   Max: 30
 
 # Do not apply the module length check to spec files.
```
sdesani, line 126: nit: might want to update the comment as well here.

beichhor: https://github.com/cerner/fhir.cerner.com/pull/419/commits/e2929f48e1c1a62380e2d66335e3ac2c0ec64403

MaxPhilips:
we aren't attaching post-merge verification here since the changes didn't affect the rendering of the site

PR #430, opened 2020-07-23
Rename the 'Practitioner' Authorization Type to 'Provider'

FFSol:
Description
----

PR Checklist
----

Screen Shot 2020-07-23 at 10 53 20 AM

### DIFF
--- a/lib/resources.rb
+++ b/lib/resources.rb
@@ -111,28 +111,16 @@ module Cerner
         end
       end
 
-      def authorization_types(provider: true, patient: false, system: false)
-        html = '<div class="auth-types">'
-        prev = false
+      def authorization_types(provider: false, patient: false, system: false)
+        template = '<a href="/authorization/#requesting-authorization-on-behalf-of-a-%s" class="%s">%s</a>'
 
-        if provider
-          html += '<a href="/authorization/#requesting-authorization-on-behalf-of-a-user" class="provider">'\
-                  'Provider</a>'
-          prev = true
-        end
+        auth_types = []
 
-        if patient
-          html += ' | ' if prev
-          html += '<a href="/authorization/#requesting-authorization-on-behalf-of-a-user" class="patient">Patient</a>'
-          prev = true
-        end
+        auth_types << template % %w[user provider Provider] if provider
+        auth_types << template % %w[user patient Patient] if patient
+        auth_types << template % %w[system system System] if system
 
-        if system
-          html += ' | ' if prev
-          html += '<a href="/authorization/#requesting-authorization-on-behalf-of-a-system" class="system">System</a>'
-        end
-
-        "#{html}</div>"
+        "<div class=\"auth-types\">#{auth_types.join(' | ')}</div>"
       end
 
       def data_currency
--- a/lib/resources.rb
+++ b/lib/resources.rb
@@ -116,9 +116,9 @@ module Cerner
 
         auth_types = []
 
-        auth_types << template % %w[user provider Provider] if provider
-        auth_types << template % %w[user patient Patient] if patient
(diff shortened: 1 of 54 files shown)


ebaranowski:
Agree

spec/lib/resources_spec.rb
```diff
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+provider# frozen_string_literal: true
```
joeratt, line 1: Copy-paste issue? I think this is breaking the tests

lib/resources.rb
```diff
@@ -108,13 +108,13 @@ def get_resource(key)
         end
       end
 
-      def authorization_types(practitioner: true, patient: false, system: false)
+      def authorization_types(provider: true, patient: false, system: false)
```
MaxPhilips, line 111: let's default provider to false if we're passing it in anyways on each method call

either that or actually take advantage of the default

lib/resources.rb
```diff
@@ -108,13 +108,13 @@ def get_resource(key)
...
         html = '<div class="auth-types">'
         prev = false
 
-        if practitioner
-          html += '<a href="/authorization/#requesting-authorization-on-behalf-of-a-user" class="practitioner">'\
-                  'Practitioner</a>'
+        if provider
```
MaxPhilips, line 115: while we're here, what do we think of beautifying this to the following:

```
      def authorization_types(provider: true, patient: false, system: false)
        template = '<a href="/authorization/#requesting-authorization-on-behalf-of-a-%s" class="%s">%s</a>'

        auth_types = []

        auth_types << template % %w[user provider Provider] if provider
        auth_types << template % %w[user patient Patient] if patient
        auth_types << template % %w[system system System] if system

        "<div class=\"auth-types\">#{auth_types.join(' | ')}</div>"
      end
```

spec/lib/resources_spec.rb
MaxPhilips: https://github.com/cerner/fhir.cerner.com/pull/430/commits/56870f0ff4573ad640d54226948788ec51129ab0#diff-0c4be351aacc6e5786906aa41d053668R1

lib/resources.rb
MaxPhilips: defaulting to false https://github.com/cerner/fhir.cerner.com/pull/430/commits/56870f0ff4573ad640d54226948788ec51129ab0#diff-7718c11ed130ac00a9854e0b089f74b4R111

lib/resources.rb
MaxPhilips: https://github.com/cerner/fhir.cerner.com/pull/430/commits/56870f0ff4573ad640d54226948788ec51129ab0#diff-7718c11ed130ac00a9854e0b089f74b4R112

lib/resources.rb
MaxPhilips: need to fix up some specs

MaxPhilips: let's add an .rspec file as well

MaxPhilips: https://github.com/cerner/fhir.cerner.com/pull/430/commits/204ced2476ff61648bb6e80f9cebee4c95f0aeab

MaxPhilips:
after rebasing, we need to switch the param name in newly added observation files

MaxPhilips:
> after rebasing, we need to switch the param name in newly added observation files

done f119e8d

lib/resources.rb
MaxPhilips: fixed lint a040879

FFSol:
Re-reviewed, re-validated.

FFSol:
Confirmed live: 
Screen Shot 2020-08-27 at 4 06 26 PM

PR #448, opened 2020-08-19
Appointment PATCH Implementation notes

Manjushavrk:
[Description](url)
----
Appointment PATCH requires both read and write scopes. This instruction is being added to the implementation notes under PATCH. 
Screenshot before merging the changes : 
Screen Shot 2020-08-18 at 2 58 38 PM

Final changes : 
Screen Shot 2020-08-24 at 8 44 40 AM

PR Checklist
----

### DIFF
--- a/content/millennium/r4/scheduling/appointment.md
+++ b/content/millennium/r4/scheduling/appointment.md
@@ -157,7 +157,7 @@ Patch an existing Appointment.
 
 _Implementation Notes_
 
-* Requires both read and write scopes.
+* For Video Visit link patch operations, both read and write scopes are required.
 * This implementation follows the [JSON PATCH](https://tools.ietf.org/html/rfc6902) spec.
 * Only operations on the paths listed below are supported.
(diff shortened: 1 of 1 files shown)


content/millennium/r4/scheduling/appointment.md
```diff
@@ -157,6 +157,7 @@ Patch an existing Appointment.
 
 _Implementation Notes_
 
+* Requires both read and write scopes.
```
kathypickering, line 160: Is both the read and write scope required for any patch or just when patching a video visit? I think it is just the Video Visit. If so, we need to clarify that. Also, I noticed we use complete sentences instead of fragments.
If so, we could say, 'For Video Visit link patch operations, both read and write scopes are required.

Manjushavrk: @kathypickering  updated here : https://github.com/cerner/fhir.cerner.com/pull/448/commits/d657dae3686d64f189fa7cde96268aa64b28a71b#diff-1d7df8ceb9f0ae8471323516e037c662R160

PR #465, opened 2020-09-18
Added Encounter patch for payment-collection-status extension

SantanuBanerjee:
Description
----

PR Checklist
----

Previously:
Screenshot 2020-09-18 at 4 57 39 PM

Change:
Screenshot 2020-09-18 at 4 56 21 PM
Screenshot 2020-09-18 at 4 55 47 PM

Updated Change:
Screen Shot 2020-09-24 at 3 07 05 PM

### DIFF
--- a/lib/resources/r4/encounter_patch.yaml
+++ b/lib/resources/r4/encounter_patch.yaml
@@ -93,7 +93,7 @@ operations:
         ]
       }
     note: >
+      Supported extensions that can be replaced:
       <ul>
-        <li>`Encounter.extension` value should be a list.</li>
-        <li>When a list is provided, the first value would be used to patch `Encounter.extension`.</li>
+        <li> <a href="https://fhir-ehr.cerner.com/r4/StructureDefinition/payment-collection-status?_format=json">Payment Collection Status</a></li>
       </ul>
--- a/lib/resources/r4/encounter_patch.yaml
+++ b/lib/resources/r4/encounter_patch.yaml
@@ -69,8 +69,8 @@ operations:
     path: /extension
     operation: replace
     type: List of Extension
-    url: http://hl7.org/fhir/encounter-definitions.html#Encounter.extension
-    description: The aray of extension of encounter.
+    url: http://hl7.org/fhir/r4/encounter-definitions.html#Encounter.extension
+    description: The array of extensions for an encounter.
     example: |
       {
         "op": "replace",
@@ -79,16 +79,16 @@ operations:
           {
             "url": "https://fhir-ehr.cerner.com/r4/StructureDefinition/payment-collection-status",
             "valueCodeableConcept": {
-            "coding": [
-              {
-                "system": "https://fhir.cerner.com/ec2458f2-1e24-41c8-b71b-0e701af7583d/codeSet/4003489",
-                "code": "40016441",
-                "display": "1st Attempt - Left Message",
-                "userSelected": true
-              }
-            ],
-            "text": "1st Attempt - Left Message"
-           }
+              "coding": [
+                {
+                  "system": "https://fhir.cerner.com/ec2458f2-1e24-41c8-b71b-0e701af7583d/codeSet/4003489",
+                  "code": "40016441",
+                  "display": "1st Attempt - Left Message",
+                  "userSelected": true
+                }
+              ],
+              "text": "1st Attempt - Left Message"
+            }
           }
         ]
       }
--- a/lib/resources/r4/encounter_patch.yaml
+++ b/lib/resources/r4/encounter_patch.yaml
@@ -64,3 +64,36 @@ operations:
         <li>`Encounter.type` value should be a list.</li>
         <li>When a list is provided, the first value would be used to patch `Encounter.type`.</li>
       </ul>
+
+  - name: replace-extension
+    path: /extension
+    operation: replace
+    type: List of Extension
+    url: http://hl7.org/fhir/encounter-definitions.html#Encounter.extension
+    description: The aray of extension of encounter.
+    example: |
+      {
+        "op": "replace",
+        "path": "/extension",
+        "value":  [
+          {
+            "url": "https://fhir-ehr.cerner.com/r4/StructureDefinition/payment-collection-status",
+            "valueCodeableConcept": {
+            "coding": [
+              {
+                "system": "https://fhir.cerner.com/ec2458f2-1e24-41c8-b71b-0e701af7583d/codeSet/4003489",
+                "code": "40016441",
+                "display": "1st Attempt - Left Message",
+                "userSelected": true
+              }
+            ],
+            "text": "1st Attempt - Left Message"
+           }
+          }
+        ]
+      }
+    note: >
+      <ul>
+        <li>`Encounter.extension` value should be a list.</li>
+        <li>When a list is provided, the first value would be used to patch `Encounter.extension`.</li>
+      </ul>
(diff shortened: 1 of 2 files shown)


lib/resources/r4/encounter_patch.yaml
```diff
@@ -64,3 +64,36 @@ operations:
...
+        "op": "replace",
+        "path": "/extension",
+        "value":  [
+          {
+            "url": "https://fhir-ehr.cerner.com/r4/StructureDefinition/payment-collection-status",
+            "valueCodeableConcept": {
+            "coding": [
+              {
+                "system": "https://fhir.cerner.com/ec2458f2-1e24-41c8-b71b-0e701af7583d/codeSet/4003489",
+                "code": "40016441",
```
sydneanschutz, line 85: This code should be valid for c1941, correct me if I'm wrong Madhur

lib/resources/r4/encounter_patch.yaml
```diff
@@ -64,3 +64,36 @@ operations:
...
+              }
+            ],
+            "text": "1st Attempt - Left Message"
+           }
+          }
+        ]
+      }
+    note: >
+      <ul>
+        <li>`Encounter.extension` value should be a list.</li>
```
sydneanschutz, line 97: Encounter.extension makes no sense here.

lib/resources/r4/encounter_patch.yaml
```diff
@@ -64,3 +64,36 @@ operations:
...
+        <li>When a list is provided, the first value would be used to patch `Encounter.extension`.</li>
```
sydneanschutz, line 98: If you have a list of two supported extensions to patch (which will be possible soon), this makes it sound like only one will be updated

lib/resources/r4/encounter_patch.yaml
```diff
@@ -64,3 +64,36 @@ operations:
         <li>`Encounter.type` value should be a list.</li>
         <li>When a list is provided, the first value would be used to patch `Encounter.type`.</li>
       </ul>
+
+  - name: replace-extension
+    path: /extension
+    operation: replace
+    type: List of Extension
+    url: http://hl7.org/fhir/encounter-definitions.html#Encounter.extension
```
sydneanschutz, line 72: http://hl7.org/fhir/r4/encounter-definitions.html#Encounter.extension

lib/resources/r4/encounter_patch.yaml
```diff
@@ -64,3 +64,36 @@ operations:
...
+    description: The aray of extension of encounter.
```
sydneanschutz, line 73: nit: 'array'
But also maybe: 'The array of extensions for an encounter'

lib/resources/r4/encounter_patch.yaml
```diff
@@ -64,3 +64,36 @@ operations:
...
+    example: |
+      {
+        "op": "replace",
+        "path": "/extension",
+        "value":  [
+          {
+            "url": "https://fhir-ehr.cerner.com/r4/StructureDefinition/payment-collection-status",
+            "valueCodeableConcept": {
+            "coding": [
```
sydneanschutz, line 82: can we fix the indentation here

lib/resources/r4/encounter_patch.yaml
SantanuBanerjee: It looks like this is the pattern followed.

lib/resources/r4/encounter_patch.yaml
SantanuBanerjee: Now, since the second extension is not patched yet,  will it sound good , if we say otherwise.
I guess it could be changed during the time when second extension patch is worked on.

lib/resources/r4/encounter_patch.yaml
SantanuBanerjee: https://github.com/cerner/fhir.cerner.com/commit/d08e9515916e1573d931f14dccab90511f189c90#diff-97d84edc87e609cdbf80d6ad20cbc681R72

lib/resources/r4/encounter_patch.yaml
SantanuBanerjee: https://github.com/cerner/fhir.cerner.com/commit/d08e9515916e1573d931f14dccab90511f189c90#diff-97d84edc87e609cdbf80d6ad20cbc681R73

lib/resources/r4/encounter_patch.yaml
SantanuBanerjee: https://github.com/cerner/fhir.cerner.com/commit/d08e9515916e1573d931f14dccab90511f189c90#diff-97d84edc87e609cdbf80d6ad20cbc681R82

lib/resources/r4/encounter_patch.yaml
sydneanschutz: I guess im thinking that because we validate single occurrence i dont know if this note is needed...?

lib/resources/r4/encounter_patch.yaml
sydneanschutz: Santanu will be removing this example

lib/resources/r4/encounter_patch.yaml
MadhurRaj: I don't think this is needed at all. The field is /extension. By the fhir spec it is defined as a list already.

lib/resources/r4/encounter_patch.yaml
MadhurRaj: We should remove this and instead have a list of supported extensions.

MadhurRaj (changes requested):

lib/resources/r4/encounter_patch.yaml
sydneanschutz: isnt type too? but its noted there....

MadhurRaj: True... I don't think it's needed there either, but let's make the change for this right now since this is new documentation.

SantanuBanerjee: https://github.com/cerner/fhir.cerner.com/commit/a9a65e45d04546e76fa09cc9d8b4ba7395d2f7d2#diff-97d84edc87e609cdbf80d6ad20cbc681R96

lib/resources/r4/encounter_patch.yaml
SantanuBanerjee: https://github.com/cerner/fhir.cerner.com/commit/a9a65e45d04546e76fa09cc9d8b4ba7395d2f7d2#diff-97d84edc87e609cdbf80d6ad20cbc681R98

SantanuBanerjee: Changed the example codevalue to point to the C1941 domain:

https://github.com/cerner/fhir.cerner.com/commit/9a21168c32e534377b2b547bec16ba3176de5ff8#diff-8915968a455f59526d7e4635365d9792R1121

https://github.com/cerner/fhir.cerner.com/commit/9a21168c32e534377b2b547bec16ba3176de5ff8#diff-97d84edc87e609cdbf80d6ad20cbc681R85

MadhurRaj (approved):

sydneanschutz (approved):

MadhurRaj:
Changes look good. Merging this.

MaxPhilips:
verified on live

Screen Shot 2020-09-24 at 5 59 12 PM

PR #504, opened 2020-12-04
Refactor site layouts

MaxPhilips:
Description
----
This PR 'anti-flattens' the layouts directory. Instead of having many files in one spot and using filenames to separate them from one another (care_provision.html and care_provision_r4.html) files now live in sorted folders. I'd like to make this change so it can be more sensible to add a contribution template to the repo (put x file in y/z/w directory, as opposed to put x_y_z file in w directory

We can also simplify some of the actual layouts in terms of removing if/else script tags

PR Checklist
----


MaxPhilips:
it's not realistic to attach screenshots for every change introduced by this PR. here's one that shows a single change working correctly

Screen Shot 2020-12-04 at 4 08 09 PM

i've compared the live site with a local site side by side on each changed page and did not notice any anomalies. additionally there are no nanoc compilation warnings for missing layout names (though this technically only proves everything that should exist does exist, not that i haven't inadvertently duplicated content in the refactor)

ebbp123 (approved):

daliboz:
@MaxPhilips since solution doesn't look like it's used anymore to auto-generate common linking/people now have to think about what "solution" they are, can you make sure you click through links and nav on the soarian side as well and ensure it lands in the right place? Also likely need to update the readme to provide updated info on how the "solution" is used now. It looks like we do still use it in some places, but not so much to make the nav and templates "solution agnostic"

MaxPhilips:
yeah, to be specific about my comment earlier, i clicked on each TOC entry for millennium dstu 2, millennium r4, and soarian dstu 2 and ensured that the same layout was pulled up in live vs. this branch

i will double check other non-TOC links we have as well

zxeleven (approved):

MaxPhilips:
I can't find any links that don't do what I want them to do

I added some info to our README with 6256ee2

daliboz:
My only concern now is that I'm curious what all the "solution" is giving us - it looks like we removed a lot of the layouts we reused between the different FHIR solutions to help keep them in sync (and some of the link generation broke down)... EG: Soarian financials and other solutions that add to our doc now need to duplicate those "overview" templates? If we're good with that, that's fine.

MaxPhilips:
what i noticed when examining the layouts dir was that a decent chunk of the currently reused layouts included conditional logic per version (e.g. if dstu 2, show x link) or were present twice (e.g. devices and devices_r4). those were the main reasons to me that a nested directory structure felt more appropriate than the current flat one

to your point - it is now more work if we decide to edit one of those layouts by changing any part of it that is still common between solutions. however, i think that this structure better shows the differences between pages that had crept into the layouts over time

MaxPhilips:
rebasing and merging

VelocityKnight:
post merge verification:
Screen Shot 2020-12-10 at 6 21 59 PM

PR #531, opened 2020-12-16
Fix Millennium Overview Table Sorting; Add CarePlan and Contract configuration

beichhor:
Description
----
Due to the way sorting of both of the arrays together was accomplished previously, the loop was prematurely ended as DSTU2 resources were completely parsed through. The alterations to the function should now properly sort through the array and reduce the potential future risks of omitting resources as we are using a `for` loop rather than a `while` loop. Resources are sorted, as the documentation states, first by the R4 resource name, if it exists, then the DSTU2 resource (if. there is no matched R4 resource).

Screen Shot 2020-12-17 at 9 08 32 AM
Screen Shot 2020-12-17 at 9 08 39 AM

PR Checklist
----

### DIFF
--- a/static/js/millennium_diff_table.js
+++ b/static/js/millennium_diff_table.js
@@ -21,6 +21,11 @@ const resourceConfig = {
       dstu2Resources: ["ProcedureRequest"],
       r4Resources: ["ServiceRequest"],
       notes: "The DSTU 2 ProcedureRequest resource was renamed to ServiceRequest in R4."
+    },
+    {
+      dstu2Resources: ["Contract"],
+      r4Resources: ["Consent"],
+      notes: "Cerner's DSTU 2 implementation of Contract resource was shifted to Consent in R4."
     }
   ],
   basicResources: [
@@ -243,8 +248,9 @@ function matchResources(conformanceData, capabilityStatementData) {
  */
 function mergeSortedArrays(includesR4, dstu2Only) {
   let result = [];
+  let includesR4Length = includesR4.length;
 
-  for(let x = 0; x < includesR4.length; x += 1) {
+  for(let x = 0; x < includesR4Length; x += 1) {
     let r4Record = includesR4[x];
 
     for (let y = 0; y < dstu2Only.length; y += 1) {
--- a/static/js/millennium_diff_table.js
+++ b/static/js/millennium_diff_table.js
@@ -25,7 +25,12 @@ const resourceConfig = {
     {
       dstu2Resources: ["Contract"],
       r4Resources: ["Consent"],
-      notes: "Cerner's DSTU 2 implementation of Contract resource was shifted to Consent in R4."
+      notes: "The resource used to represent people who are authorized to view a patient's data shifted from DSTU2 Contract to R4 Consent."
+    },
+    {
+      dstu2Resources: ["CarePlan"],
+      r4Resources: ["CarePlan", "CareTeam"],
+      notes: "DSTU2 CarePlan was split into R4 CarePlan and CareTeam."
     }
   ],
   basicResources: [
--- a/static/js/millennium_diff_table.js
+++ b/static/js/millennium_diff_table.js
@@ -235,38 +235,29 @@ function matchResources(conformanceData, capabilityStatementData) {
 }
 
 /**
- * Merges the two sorted arrays alphabetically.
+ * Merges the two sorted arrays alphabetically. Entries are ordered by their R4 name first, if no R4 resource
+ * exists then the DSTU2 is used.
  * @param {array} includesR4 - The array of resource objects that contain R4 resources.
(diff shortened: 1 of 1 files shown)


ericsornson (approved):

ericsornson (approved):
:shipit:

ebbp123 (approved):

static/js/millennium_diff_table.js
```diff
@@ -235,38 +235,29 @@ function matchResources(conformanceData, capabilityStatementData) {
...
+ * Merges the two sorted arrays alphabetically. Entries are ordered by their R4 name first, if no R4 resource
+ * exists then the DSTU2 is used.
  * @param {array} includesR4 - The array of resource objects that contain R4 resources.
  * @param {array} dstu2Only - The array of resource objects that only contain DSTU2 resources.
  * @returns {array} The resultant array after merging.
  */
 function mergeSortedArrays(includesR4, dstu2Only) {
   let result = [];
-  let i = 0;
-  let j = 0;
```
DanielWhiteCerner, line 246: nit: Defining the lets outside the loop like it was before is a JavaScript standard.

static/js/millennium_diff_table.js
```diff
@@ -235,38 +235,29 @@ function matchResources(conformanceData, capabilityStatementData) {
...
-        i += 1;
-      }
-      else {
-        result.push(dstu2Only[j]);
-        j += 1;
+
+  for(let x = 0; x < includesR4.length; x += 1) {
+    let r4Record = includesR4[x];
+
+    for (let y = 0; y < dstu2Only.length; y += 1) {
```
DanielWhiteCerner, line 250: nit: You'll speed up the search a bit if you define variables for `includesR4.length` and `dstu2Only.length` before starting the loop as well. The current syntax is going to check the length each time you go through the loop.

static/js/millennium_diff_table.js
```diff
@@ -235,38 +235,29 @@ function matchResources(conformanceData, capabilityStatementData) {
...
+      if (r4Record.r4Resources[0].resourceName > dstu2Only[y].dstu2Resources[0].resourceName) {
```
DanielWhiteCerner, line 251: It's an unfortunate requirement of JavaScript without `dig`, but you should probably keep the `hasOwnProperty` checks for `r4Resources` and `dstu2Resources`.

DanielWhiteCerner (changes requested):

static/js/millennium_diff_table.js
beichhor: dstu2Only.length needs to stay as is since we are splicing the array and removing an element

static/js/millennium_diff_table.js
beichhor: The hasOwnProperty, previously, was trying to perform a check to see if the R4 resources also contained a matching DSTU2 record, this is no longer how the sorting is performed and these arrays and hashes are internally built. We know that the r4Record will always have a resourceName and that dstu2Only will have a dstu2Resource name

static/js/millennium_diff_table.js
beichhor: `let`'s are block scoped. Meaning by definition, if we defined them at the top of the function, they will be available outside of the loops (where they are only needed). Given that, previously, it was a `where` loop, versus now a `for` loop. And I can't find anything in eslint regarding this or anything stating this for a `for` loop in general, this seems unnecessary (and potentially introduce more bugs).

static/js/millennium_diff_table.js
DanielWhiteCerner: Found that code.

static/js/millennium_diff_table.js
DanielWhiteCerner: Iterating on a changing array seems like a bad idea. It would be better to use another variable as a starting point for "y", but this is fine.

DanielWhiteCerner (approved):

static/js/millennium_diff_table.js
DanielWhiteCerner: _All_ variables in JavaScript are defined at the top of the method. That's why my personal preference for any JavaScript is to define all variables at the top to remind readers of that.

The new ES6 syntax for `let` just avoids hoisting.

static/js/millennium_diff_table.js
beichhor: https://github.com/cerner/fhir.cerner.com/commit/e52ecc574fd6f521216755e0410a2911dd8ddb5a

ericsornson (approved):

DanielWhiteCerner (approved):

ebbp123 (approved):
Looks good in firefox as well.

ebbp123:
Change is live: 
Screen Shot 2020-12-17 at 4 01 29 PM
Screen Shot 2020-12-17 at 4 01 54 PM

PR #561, opened 2021-03-01
Updated related person remove telecom operation

Nandita92:
Description
----
Added remove operation for telecom field in related person patch capability.

PR Checklist
----
image
image

### DIFF
--- a/lib/resources/r4/related_person_patch.yaml
+++ b/lib/resources/r4/related_person_patch.yaml
@@ -302,3 +302,19 @@ operations:
         <li>If no suffix value is provided, the suffix on the name will be unset.</li>
         <li>Only one suffix value may be provided.</li>
       </ul>
+
+  - name: remove-telecom
+      path: /telecom/{index}
+      operation: remove
+      type: N/A
+      url: https://hl7.org/fhir/R4/patient-definitions.html#related person.telecom
+      description: Remove the telecom at the given <code>{index}</code> in the list of related person telecoms.
+      example: |
+        {
+          "op": "remove"
+          "path": "/telecom/1"
+        }
+      note: >
+        <ul>
+          <li>Requires a <a href="#test-telecom-id">test operation</a> to be provided for the `ContactPoint.id` of the telecom intended to be removed.</li>
+        </ul>
--- a/lib/resources/r4/related_person_patch.yaml
+++ b/lib/resources/r4/related_person_patch.yaml
@@ -205,6 +205,22 @@ operations:
         "value": "CI-PH-2807950-0"
       }
 
+  - name: remove-telecom
+    path: /telecom/{index}
+    operation: remove
+    type: N/A
+    url: https://hl7.org/fhir/R4/relatedperson-definitions.html#RelatedPerson.telecom
+    description: Remove the telecom at the given <code>{index}</code> in the list of related person telecoms.
+    example: |
+      {
+        "op": "remove"
+        "path": "/telecom/0"
+      }
+    note: >
+      <ul>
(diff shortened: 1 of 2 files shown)


lib/resources/r4/related_person_patch.yaml
```diff
@@ -159,3 +159,32 @@ operations:
...
+        "op": "test",
+        "path": "/telecom/0/id",
+        "value": "CI-PH-2807950-0"
+      }
+
+  - name: remove-telecom
+    path: /telecom/{index}
+    operation: remove
+    type: N/A
+    url: https://hl7.org/fhir/R4/patient-definitions.html#related person.telecom
```
SantanuBanerjee, line 180: I think the url will not be working. First of all patient-definitions will take to the patient page, which is wrong and then #related person will not point to any specific field.
So, it should ideally be:  https://hl7.org/fhir/R4/relatedperson-definitions.html#RelatedPerson.telecom

lib/resources/r4/related_person_patch.yaml
```diff
@@ -159,3 +159,32 @@ operations:
...
+    description: Remove the telecom at the given <code>{index}</code> in the list of related person telecoms.
+    example: |
+      {
+        "op": "remove"
+        "path": "/telecom/1"
```
SantanuBanerjee, line 185: I think, since the other examples here are pointing to the 0 index, we can use the same index here, to keep things in sync.

lib/resources/r4/related_person_patch.yaml
Nandita92: Updated here: https://github.com/cerner/fhir.cerner.com/commit/150edb0c78f14b75bf4da4f1412e24b65b90c50a

lib/resources/r4/related_person_patch.yaml
Nandita92: Updated here: https://github.com/cerner/fhir.cerner.com/commit/150edb0c78f14b75bf4da4f1412e24b65b90c50a

lib/resources/r4/related_person_patch.yaml
SantanuBanerjee: Now when I recheck I think the link I provided was also incorrect, sorry for that.
It should be: https://hl7.org/fhir/R4/relatedperson-definitions.html#RelatedPerson.telecom

Please check the link once to see if it is taking you to the page, before you push it.

Nandita92: Updated the url: https://github.com/cerner/fhir.cerner.com/commit/8c298287e4acd1221bb0bdd92719a25868a8496b

SantanuBanerjee (approved):

exopsRamba (approved):

alexbeckner1 (approved):

sujanb12 (approved):

lib/resources/r4/related_person_patch.yaml
```diff
@@ -289,3 +289,19 @@ operations:
         <li>If no suffix value is provided, the suffix on the name will be unset.</li>
         <li>Only one suffix value may be provided.</li>
       </ul>
+
+  - name: remove-telecom
```
ebbp123, line 293: Can we move this to be under `add-telecom`? I think the order could be add, test, remove.

ebbp123: Will need to update r4_examples_relatedperson.rb to match the new order.

Nandita92: Updated here: https://github.com/cerner/fhir.cerner.com/commit/e044c496534353b5646dd08dae7a0c036e2b6dc4

ebbp123 (approved):

BS062721:
Verified on live
Screen Shot 2021-03-18 at 2 48 57 PM

PR #701, opened 2021-07-15
Condition Category Classification Code

sydneanschutz:
Description
----
Return Condition Category Classification Code

PR Checklist
----

Screen Shot 2021-07-15 at 12 41 13 PM
Screen Shot 2021-07-15 at 12 41 38 PM
Screen Shot 2021-07-15 at 12 41 50 PM
Screen Shot 2021-07-15 at 12 42 05 PM

### DIFF
--- a/lib/resources/example_json/r4_examples_condition.rb
+++ b/lib/resources/example_json/r4_examples_condition.rb
@@ -238,9 +238,8 @@ module Cerner
         'status': 'generated',
         'div': '<div xmlns="http://www.w3.org/1999/xhtml"><p><b>Condition</b></p><p><b>Patient</b>: '\
               'SMART, TIMMY</p><p><b>Health Concern</b>: THIS IS A TEST GOAL, THIS IS A TEST GOAL</p>'\
-              '<p><b>Category Classification</b>: Medical</p><p><b>Clinical Status</b>: Active</p>'\
-              '<p><b>Verification Status</b>: Confirmed</p><p><b>Onset</b>: Dec  3, 2020</p>'\
-              '<p><b>Recorded Date</b>: Dec  3, 2020 10:47 P.M. UTC</p></div>'
+              '<p><b>Clinical Status</b>: Active</p><p><b>Verification Status</b>: Confirmed</p><p>'\
+              '<b>Onset</b>: Dec  3, 2020</p><p><b>Recorded Date</b>: Dec  3, 2020 10:47 P.M. UTC</p></div>'
       },
       'clinicalStatus': {
         'coding': [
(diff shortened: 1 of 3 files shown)


lib/resources/r4/condition.yaml
```diff
@@ -167,14 +168,21 @@ fields:
     }
   binding:
     description: A category assigned to the condition.
```
RayGunY, line 170: On the jira it says to update this to "A classification assigned......" is that what we want or did we change our minds to keep it as category?

sydneanschutz: this description is for the field category, like in general. That description is for specifically the classification code. not sure where i would put it

lib/resources/example_json/r4_examples_condition.rb
```diff
@@ -237,6 +274,17 @@ module Resources
             }
           ],
           'text': 'Health Concern'
+        },
+        {
```
MadhurRaj, line 278: health concerns won't the classification.

lib/resources/example_json/r4_examples_condition.rb
```diff
@@ -384,6 +443,17 @@ module Resources
               }
             ],
             'text': 'Health Concern'
+          },
+          {
```
MadhurRaj, line 447: same here.

MadhurRaj (changes requested):

lib/resources/example_json/r4_examples_condition.rb
sydneanschutz: https://github.com/cerner/fhir.cerner.com/pull/701/commits/b8983bad284d00a1972186571002ee91b6559052#diff-20bf74979e847ab2cdb40b65f16e39069096d8a3a39fefe73adeae819dc6bbdfL241

lib/resources/example_json/r4_examples_condition.rb
sydneanschutz: https://github.com/cerner/fhir.cerner.com/pull/701/commits/b8983bad284d00a1972186571002ee91b6559052#diff-20bf74979e847ab2cdb40b65f16e39069096d8a3a39fefe73adeae819dc6bbdfL411

lib/resources/r4/condition.yaml
RayGunY: Ok fair enough then, not too worried about it

RayGunY (approved):

MadhurRaj (approved):

sydneanschutz:
Verified on live
Screen Shot 2021-07-26 at 11 50 29 AM

PR #721, opened 2021-08-11
Immunization target-disease support

lalitha94:
Description
----
Added support for Immunization search by target-disease.
Exposing Immunization.protocolApplied field

PR Checklist
----
- [X] Screenshot(s) of changes attached before changes merged.

Screen Shot 2021-08-11 at 8 58 12 AM
Screen Shot 2021-08-11 at 8 58 29 AM
Screen Shot 2021-08-11 at 8 58 43 AM

### DIFF
--- a/lib/resources/r4/immunization.yaml
+++ b/lib/resources/r4/immunization.yaml
@@ -414,3 +414,23 @@ fields:
     Will be set to true if not specified.
     For CREATE action should not be set to true when reportOrigin is populated.
     For UPDATE action should not be set to true.
+
+- name: protocolApplied.targetDisease
+  required: 'No'
+  type: CodeableConcept
+  url: https://www.hl7.org/fhir/r4/immunization-definitions.html#Immunization.protocolApplied.targetDisease
+  description: The vaccine preventable disease the dose is being administered against.
+  action: terminology
+  binding:
+    description: The vaccine preventable disease the dose is being administered against.
+    terminology:
+    - display: CVX
+      system: http://hl7.org/fhir/sid/cvx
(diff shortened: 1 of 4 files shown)


joeratt (approved):

fpivonka (approved):

lib/resources/r4/immunization.yaml
```diff
@@ -414,3 +414,23 @@ fields:
...
+  action: terminology
+  binding:
+    description: The vaccine preventable disease the dose is being administered against.
+    terminology:
+    - display: CVX
+      system: http://hl7.org/fhir/sid/cvx
+      info_link: http://hl7.org/fhir/r4/cvx.html
+    - display: SNOMED CT
+      system: http://snomed.info/sct
+      info_link: http://hl7.org/fhir/r4/snomedct.html
```
Mohammad-Isyroqi-Fathan, line 432: Do we actually have SNOMED mapping for this?

lalitha94: Thanks for pointing it out. I think currently we do not have mappings but we do want to support it. @fpivonka can we leave it or should we add it once the mappings are completed?

fpivonka: @lalitha94 @Mohammad-Isyroqi-Fathan I think we can leave it off for now. We do support the scope for SNOMED CT but we have not mapped clients that way.

ashu-bdvt:
Post deployment screenshots:

Screenshot 2021-09-09 at 8 02 30 PM
Screenshot 2021-09-09 at 8 00 49 PM
Screenshot 2021-09-09 at 8 00 24 PM

PR #777, opened 2021-11-29
Clarify token revocation behavior

mbeerman:
Description
----
In the course of testing related to the 21st Century Cures Act (using the ONC's "Inferno" test suite), some questions arose related to the Authorization Server's handling of token revocation; these changes attempt to clarify those behaviors.

PR Checklist
----
- [X] Screenshot(s) of changes attached before changes merged.

Screen Shot 2021-11-29 at 5 00 50 PM

### DIFF
--- a/content/authorization.md
+++ b/content/authorization.md
@@ -1058,11 +1058,9 @@ preceding documentation.
   behalf of a user?
 
 > Cerner currently does not have a mechanism that allows
-  client applications to revoke refresh tokens.  Cerner is
-  currently tracking the progress of the IETF Proposed
-  Standard RFC
-  ["OAuth 2.0 Token Revocation"](https://tools.ietf.org/html/rfc7009)
-  for further evaluation of such capabilities.
+  client applications to revoke refresh tokens.  If/when
+  such functionality is implemented, it will follow RFC 7009
+  ["OAuth 2.0 Token Revocation"](https://tools.ietf.org/html/rfc7009).
 
 - How can my application participate in log out
   mechanisms provided by the organization's single
--- a/content/authorization.md
+++ b/content/authorization.md
@@ -1054,6 +1054,18 @@ preceding documentation.
   such devices to participate in the authorization
   ecosystem.
 
+- What happens when a user revokes my application's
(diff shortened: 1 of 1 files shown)


mbeerman:
I will of course squash these changes when merging. :)

jfrhong (approved):

content/authorization.md
```diff
@@ -1054,15 +1054,30 @@ preceding documentation.
   such devices to participate in the authorization
   ecosystem.
 
+- What happens when a user revokes my application's
+  access to their data?
+
+> When a user revokes an application's access, its
+  *refresh* tokens immediately become non-functional.
```
kolkheang, line 1061:
```
*refresh* tokens immediately
```
Are there multiple refresh tokens? Or, will there only be one?

Wondering if we can drop the **s**.

Same question applies to the one below as well.

```
*Access* tokens cannot be 
```

content/authorization.md
```diff
@@ -1054,15 +1054,30 @@ preceding documentation.
   such devices to participate in the authorization
   ecosystem.
 
+- What happens when a user revokes my application's
+  access to their data?
+
```
kolkheang, line 1059: I am guessing that by having this section, it's implied that there is a way to revoke the refresh token for an app. Shall we mention that the operation can be done by an administrator currently?

kolkheang, line 1059: Shall we mention that a refresh token can be revoked at anytime and that it's an expected behavior and the app needs to handle the scenario properly?

content/authorization.md
mbeerman: I think an application could have several in theory (though it's probably rare in practice).

content/authorization.md
mbeerman: It *can't* be done by an administrator currently (only by the users themselves).

mbeerman: Maybe, but there's some risk of simply restating the OAuth RFCs (which I try to avoid when possible).

aki-mathur:
Screenshot 2021-12-02 at 6 37 42 PM

PR #845, opened 2022-05-19
Updating a few references to which domain should have the TXT records

brailsmt:
Description
----

PR Checklist
----
- [ ] Screenshot(s) of changes attached before changes merged.


scott-bach (approved):
+1 these look good.

brailsmt:
How should I attach screenshots?  It would need to be 4 images (2 each for each screen).  I'll go ahead and do it if needed, but a diff should provide a much better overview of the changes than a screenshot.  The before can be seen at the cerner:main branch and the after can be seen at brailsmt:main.  Please advise.

mbeerman (approved):

PR #853, opened 2022-06-13
Remove provider offline access guidance

zxeleven:
Description
----
Remove the section about provider offline access from the Authorization FAQ.

Screenshots
----
Before:
Screen Shot 2022-06-13 at 12 11 13 PM

After:
Screen Shot 2022-06-13 at 12 10 39 PM

PR Checklist
----

### DIFF
--- a/content/authorization.md
+++ b/content/authorization.md
@@ -1176,11 +1176,6 @@ Disable or rotate your system account via Cerner Central as
   soon as possible via
   [Cerner Central System Account Management][SYSTEM-ACCOUNTS].  
 
-**Is offline_access supported for healthcare providers?**
-
-Cerner does not currently have support for offline_access
-  for providers.  It is currently under consideration.
-
 **How can I deliver my solution to users utilizing its own
   Citrix instance with XenApp?**
 


content/authorization.md
```diff
@@ -1176,11 +1176,6 @@ Disable or rotate your system account via Cerner Central as
   soon as possible via
   [Cerner Central System Account Management][SYSTEM-ACCOUNTS].  
 
-**Is offline_access supported for healthcare providers?**
```
markdcummings, line 1179: This may have been discussed somewhere already (though I didn't see it on the JIRA), but is there value in keeping this section but updating it to say that it is supported? Thinking of someone who might have previously read this FAQ and wondering if this should be updated to something like "Cerner now supports offline_access for providers" with maybe additional useful information ("as of <version or date").

markdcummings: After some offline discussion it sounds like removing this item is the desired course of action.

markdcummings (approved):

chenyixuan625 (approved):

PR #877, opened 2022-08-22
Add information about scheme flooding

mbeerman:
Description
----
Microsoft Edge added new security measures in March 2022 that can be problematic for developers of native SMART applications; document what we know about the causes and resolutions.

PR Checklist
----

Screen Shot 2022-08-22 at 12 49 35 PM

### DIFF
--- a/content/authorization.md
+++ b/content/authorization.md
@@ -653,6 +653,8 @@ in turn determine which open instance of the application
 (if multiples are allowed) initiated the equest via
 examination of the "state" parameter.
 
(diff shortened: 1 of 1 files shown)


whitehatguy (approved):
+1

scott-bach:
Can you add the cause or conditions in which they would need to have these settings required or, if they are always required, let them know that also?

scott-bach (approved):
Looks good with the update to include the error they might be seeing.

aki-mathur:
Screenshot 2022-08-26 at 12 48 39 PM

PR #903, opened 2022-09-21
Remove Group from resouces table and fix health cards link

ebbp123:
Description
----
Remove the Group resource from the Millennium Resources Table.
Fix Health Cards link in Patient resource documentation.

Dev Validation
----
Screen Shot 2022-09-21 at 2 52 12 PM
Screen Shot 2022-09-21 at 2 52 35 PM

PR Checklist
----

### DIFF
--- a/content/millennium/r4/base/individuals/patient.md
+++ b/content/millennium/r4/base/individuals/patient.md
@@ -323,4 +323,4 @@ See [Health Cards] documentation for more details about this operation.
 [US Core Ethnicity]: https://hl7.org/fhir/us/core/StructureDefinition-us-core-ethnicity.html
 [US Core Birth Sex]: https://hl7.org/fhir/us/core/StructureDefinition-us-core-birthsex.html
 [Communication Preference]: https://fhir-ehr.cerner.com/r4/StructureDefinition/communication-preference?_format=json
-[Health Cards]: /millennium/r4/other/health-cards/
+[Health Cards]: /millennium/r4/foundation/other/health-cards/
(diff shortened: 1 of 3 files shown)


amcginn:
since resourceObject.interaction is 0..*, shouldn't we still check if it exists?

ebbp123:
> since resourceObject.interaction is 0..*, shouldn't we still check if it exists?

So I don't think we will run into another resource that has no interactions. In the case of Group, we added it to the CapabilityStatement to expose the group-export OperationDefinition in the metadata, and it has no interactions since this is not a "resource" that our clients will be able to interact with directly.

Let me know if I should add the check regardless.

amcginn:
> > since resourceObject.interaction is 0..*, shouldn't we still check if it exists?
> 
> So I don't think we will run into another resource that has no interactions. In the case of Group, we added it to the CapabilityStatement to expose the group-export OperationDefinition in the metadata, and it has no interactions since this is not a "resource" that our clients will be able to interact with directly.
> 
> Let me know if I should add the check regardless.

I don't know that I can answer that beyond noting the API spec definition

ebbp123:
> > > since resourceObject.interaction is 0..*, shouldn't we still check if it exists?
> > 
> > 
> > So I don't think we will run into another resource that has no interactions. In the case of Group, we added it to the CapabilityStatement to expose the group-export OperationDefinition in the metadata, and it has no interactions since this is not a "resource" that our clients will be able to interact with directly.
> > Let me know if I should add the check regardless.
> 
> I don't know that I can answer that beyond noting the API spec definition

I added the check since as you say in the spec, it can be null: https://github.com/cerner/fhir.cerner.com/pull/903/commits/c6ada6c2a986d9ac1299c1389e53d65f1aff4244.
Since interaction.code is 1..1 we should not need to worry about it being null.

amcginn (approved):

SA030800 (approved):

PR #906, opened 2022-09-28
Fix request headers links for bulk-data

ebbp123:
Description
----
Fix the path used for the links for the request headers

Local Validation
----
Screen Shot 2022-09-29 at 11 55 35 AM

PR Checklist
----

### DIFF
--- a/lib/resources.rb
+++ b/lib/resources.rb
@@ -27,19 +27,20 @@ module Cerner
         502 => '502 Bad Gateway'
       }.freeze
 
-      def headers(status: nil, head: {}, fhir_json: false)
+      def headers(status: nil, head: {}, fhir_json: false, bulk: false, r4: true)
         lines = status ? ["Status: #{STATUSES[status]}"] : []
+        path = headers_path(bulk, r4)
 
         head.each do |key, value|
           lines << case key
                    when :pagination
                      link_header(value)
                    when :Accept
-                     "<a href=\"../../#media-types\">#{key}</a>: #{value}"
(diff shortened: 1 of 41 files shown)


lib/resources.rb
```diff
@@ -64,6 +65,13 @@ def link_header(rels)
         lines
       end
 
+      def headers_path(bulk, r_4)
```
SA030800, line 68: nit: r4

lib/resources.rb
```diff
@@ -64,6 +65,13 @@ def link_header(rels)
...
+        path = r_4 ? '../../../' : '../../'
+        path = bulk ? '../../bulk-data' : path
```
SA030800, line 70: do we need to have the bulk-data hard coded can't we use the relative path?

ebbp123: rubocop throws the following warning since it is a numbered variable, hence why I added the underscore:
`Naming/VariableNumber: Use snake_case for variable numbers.`

lib/resources.rb
ebbp123: rubocop throws the following warning since it is a numbered variable, hence why I added the underscore:
`Naming/VariableNumber: Use snake_case for variable numbers.`

SA030800 (approved):

MadhurRaj (approved):

ebbp123:
Post-Validation
----
Screen Shot 2022-09-29 at 4 17 12 PM
Screen Shot 2022-09-29 at 4 17 05 PM

PR #985, opened 2023-09-20
Review and Corrections for Device

RicardoCrespo-ora:
Description
----
In preparation for migrating to Oracle docs, we are reviewing and correcting any inconsistencies, typos, broken links, etc. This PR aims to correct any of the above that needs to be addressed within Consent/Contract and any other information directly referenced by it.

### Modified implementation notes to match styling of other pages

**Before**
image

**After**
image

---

### Search request and response was being showcased twice, removed the duplicate

**Before**
image

**After**
image

---

### Duplicated Response/Request for /Device/:id 

**Before**
image
image

**After**
image
image

---

### Minor formatting to error example title for readability

**Before**
image

**After**
image

PR Checklist
----
- [x] Screenshot(s) of changes attached after changes merged and published.

### DIFF
--- a/content/millennium/r4/base/entities/device.md
+++ b/content/millennium/r4/base/entities/device.md
@@ -60,9 +60,10 @@ Search for Devices that meet supplied query parameters:
 
 _Implementation notes_
 
-* The `_revinclude` parameter may be provided once with the value `Provenance:target`. Example: `_revinclude=Provenance:target`
-
-* The `_revinclude` parameter may be provided with the `_id/patient` parameter. Example: `_id=214938095&_revinclude=Provenance:target`
+* The `_revinclude` parameter 
+    * May be provided once with the value `Provenance:target`. Example: `_revinclude=Provenance:target`
+    * May be provided with the `_id` parameter. Example: `_id=214938095&_revinclude=Provenance:target`
+    * May be provided with the `patient` parameter. Example: `patient=12742397&_revinclude=Provenance:target`
 
 * When `_revinclude` is provided in a request to the closed endpoint, the OAuth2 token must include the `user/Provenance.read` scope. Currently `patient/Provenance.read` is not supported and hence `_revinclude` cannot be utilised for patient persona.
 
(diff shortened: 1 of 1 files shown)


kckeim (changes requested):
The `_id/patient` section should actually be broken apart on line 65
It's two separate inputs, so, it could be "provided with the `_id` or `patient` parameters" instead of as-written.

image

*******************************
I like these other changes, especially since we're removing a lot of duplicated / unnecessary content.  I'm not sure if we want to keep the different "kinds" of authorizations' requests in there, since the request is functionally the same, and retrieves the same information either way, unless Aaron or someone else sees it as necessary to keep in there.

kckeim (approved):

content/millennium/r4/base/entities/device.md
```diff
@@ -56,35 +56,25 @@ Search for Devices that meet supplied query parameters:
 --------------|-------------------|---------------|------------------------------------------------------------------------
  `_id`        | This or `patient` | [`token`]     | The logical resource id associated with the Device. Example: `7890`
  `patient`    | This or `_id`     | [`reference`] | The patient to whom the device is affixed. Example: `12345`
-`_revinclude` | N                 | [`token`]     | Provenance resource entries to be returned as part of the bundle. Example: `_revinclude=Provenance:target`
+`_revinclude` | No                | [`token`]     | Provenance resource entries to be returned as part of the bundle. Example: `_revinclude=Provenance:target`
```
amcginn, line 59: I can't remember, does this match the other ones we've done so far?

amcginn:
the duplicate content has bugged me for so long, thank you 🥲

amcginn (approved):

content/millennium/r4/base/entities/device.md
kckeim: I think I've been doing "N"/"Y" on most.

For what it's worth, I expected that these sections would need to get ripped apart when we get new formatting guidelines, so didn't have it as a high priority.

RicardoCrespo-ora: Reverted it back to N

kckeim (approved):

balanionut-cerner (approved):

